PR #3, opened 2017-07-24
Minor correction to documentation

LeahPrice:
Correcting the description of GetParticleWeight().

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,8 @@
 2017-07-24  Leah South  <leah.south@hdr.qut.edu.au>
+
+    * inst/include/sampler.h: Correcting the description of GetParticleWeight().
+    
+2017-07-24  Leah South  <leah.south@hdr.qut.edu.au>
     
     * src/init.c: Removed due to use of Rcpp attributes.
     * src/RcppExports.cpp: Initial creation with Rcpp attributes.
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,41 +1,36 @@
 2017-07-24  Leah South  <leah.south@hdr.qut.edu.au>
     
+    * inst/include/particle.h: Removed (no longer needed now with population.h).
+    
     * src/init.c: Removed due to use of Rcpp attributes.
     * src/RcppExports.cpp: Initial creation with Rcpp attributes.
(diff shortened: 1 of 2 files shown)


ChangeLog
```diff
@@ -1,4 +1,8 @@
 2017-07-24  Leah South  <leah.south@hdr.qut.edu.au>
+
+    * inst/include/sampler.h: Correcting the description of GetParticleWeight().
+    
+2017-07-24  Leah South  <leah.south@hdr.qut.edu.au>
```
eddelbuettel: Nitwick: when one has two entries _with the same data and author_ one tends to merge those.

But thanks for keeping up with ChangeLog!

LeahPrice: Sorry are you referring to merging the entries in the ChangeLog or another type of merge? I just did a commit to remove the unnecessary ChangeLog entry.

eddelbuettel:
@LeahPrice You removed the entire ChangeLog entry.  That is not what I suggested. I suggested to regroup all entries having the same date under the same date.  Anyway....

LeahPrice:
@eddelbuettel Okay, thanks for clarifying that. When you said the same data and author, I misinterpreted it as the same author and general theme of the update. I removed the extra entry because it was just an error introduced in the previous change.

I've fixed this up now and combined into one.

eddelbuettel:
Merged.  I still think you need entries for your name and email in `~/.gitconfig`:

```sh
edd@brad:~$ head -3 .gitconfig
[user]
        name = Dirk Eddelbuettel
        email = edd@debian.org
edd@brad:~$
```

so that your commits are market as yours.  Just about every git beginner's tutorial mentions this, you may have skipped it.  Might be worth adding (though again I have no idea where to do this on Windows, but lots of other users seem to manage....)

LeahPrice:
Thanks @eddelbuettel for picking up that it didn't work. Could you please let me know where you found that quoted log that said <unknown> so I can follow it up? The Gitkraken tutorial says it happens automatically - updating my profile didn't work but surely the manual fix will.

eddelbuettel:
`git log` on the command shell.  I use an alias that pretty-prints the log output (and there are many).

PR #5, opened 2017-07-26
Removing Multinomial helper function

LeahPrice:
Removing the Multinomial helper function and changing uRSCount to type arma::Col<int>.

### DIFF
--- a/inst/include/sampler.h
+++ b/inst/include/sampler.h
@@ -71,7 +71,7 @@ namespace smc {
         ///Structure used internally for resampling.
         arma::vec dRSWeights;
         ///Structure used internally for resampling.
-        arma::Col<unsigned int> uRSCount;
+        arma::Col<int> uRSCount;
         ///Structure used internally for resampling.
         arma::Col<unsigned int> uRSIndices;
 
@@ -151,8 +151,6 @@ namespace smc {
         sampler(const sampler<Space> & sFrom);
         ///Duplication of smc::sampler is not currently permitted.
         sampler<Space> & operator=(const sampler<Space> & sFrom);
-        ///Generate a multinomial random vector with parameters (n,w[1:k]) and store it in X
-		void Multinomial(unsigned n, unsigned k, arma::vec w, unsigned int * X);		
 		
     };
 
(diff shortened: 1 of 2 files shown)


inst/include/sampler.h
```diff
@@ -71,7 +71,7 @@ namespace smc {
         ///Structure used internally for resampling.
         arma::vec dRSWeights;
         ///Structure used internally for resampling.
-        arma::Col<unsigned int> uRSCount;
+        arma::Col<int> uRSCount;
```
adamjohansen: A trivial detail, but the u prefix denotes an unsigned value...

Now this is based on some pretty old code and many don't like Hungarian notation any more. Which is fine, it's not something I'm religious about, but if we use it or something that looks like it then it should be meaningful. nRSCount would better match an object of this sort, but as soon as we're using arma:Col rather than a simple array even that's a bit of a stretch.

With that in mind, it would be good to think about how best to refer to name this one object to avoid introducing any unnecessary confusion and at some later stage (not necessarily within GSOC and this isn't your problem so much as an RcppSMC one) to think a bit about variable naming conventions.

adamjohansen:
This looks fine to me -- modulo a slightly pedantic quibble over variable naming which is largely a hangover from my own slightly arcane approach to naming variables in an earlier project -- thanks for the quick update...

inst/include/sampler.h
LeahPrice: Thanks, that's good to know. Unfortunately I don't think this is the only variable name I've done this with.

adamjohansen: I wouldn't worry too much, I think most people would curse me more for introducing the convention in the first place. It would be good to try at least to be consistent within the code, but there's absolutely no urgency.

eddelbuettel:
> I don't think this is the only variable name I've done this with.

In general, it is common practice to NOT change other's people's preferred indentation, naming (camelCase versus snake_case versus hungarian versus....) if that is the sole change.

It's better to focus on the big picture.  I'll open a ticket so that we can set up a better plan.

Also: `arma::Col<unsigned>` is IIRC the same as `arma::uvec` which is more concise.  We generally don't use the explicitly templated form.

LeahPrice:
Yes I didn't intentionally change those things as I don't know enough about them to have a preference (and especially didn't when first starting!). I will focus on the things I've put in the plan, but it should be very quick to fix up that latter comment (maybe along with one of my upcoming pull requests).

PR #7, opened 2017-07-27
Log Ratio of Normalizing Constants

LeahPrice:
Adding estimators of the log ratio of normalizing constant based on
- the standard SMC estimator
- path sampling estimator with 1st and 2nd order trapezoidal integration

### DIFF
--- a/inst/include/population.h
+++ b/inst/include/population.h
@@ -37,6 +37,10 @@
 #include <cmath>
 
 namespace smc {
+
+    /// A stable calculation of the log sum of the weights, used in ESS calculations
+    double stableLogSumWeights(const arma::vec & logw);
+    
     /// A template class for the particles of an SMC algorithm
     template <class Space> class population
     {
@@ -135,6 +139,17 @@ namespace smc {
         this->logweight = pFrom.logweight;
 
         return *this;
+    }
+    
+    /// This function performs a stable calculation of the log sum of the weights, which is useful for
+    /// normalising weights, calculating the effective sample size and estimating the normalising constant.
+    ///
+    /// \param logw The log weights of interest.
+    inline double stableLogSumWeights(const arma::vec & logw){
+        long N = logw.n_rows;
+        double dMaxWeight = arma::max(logw);
+        double sum = arma::sum(exp(logw - dMaxWeight*arma::ones(N)));
+        return (dMaxWeight + log(sum));
     }
 }
 
--- a/inst/include/population.h
+++ b/inst/include/population.h
@@ -140,17 +140,6 @@ namespace smc {
 
         return *this;
     }
-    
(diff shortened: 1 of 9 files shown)


LeahPrice:
I moved the function `stableLogSumWeights` to the `population` header file. It compiled for me and I didn't get any warnings, but I didn't get any warnings before Dirk added the forward declaration either. I think it's okay since Travis CI checks passed.

Some of this is a little different to what I have in my master copy - I changed the template parameter for history to make it easier for users to write the path sampling integrands. I've checked that everything still works properly on the linear regression example but I'm saving that for a future pull request where I'll document it properly. Please let me know if you'd like me to add it in now.

inst/include/population.h
```diff
@@ -136,6 +140,17 @@ namespace smc {
 
         return *this;
     }
+    
+    /// This function performs a stable calculation of the log sum of the weights, which is useful for
+    /// normalising weights, calculating the effective sample size and estimating the normalising constant.
+    ///
+    /// \param logw The log weights of interest.
+    inline double stableLogSumWeights(const arma::vec & logw){
```
adamjohansen: I'm doing this too quickly, because I need to check out of my accommodation shortly, but is this "code in a header"? With the exception of /templated/ functions which are instructions to the compiler to generate code rather than code itself, or inline member functions, it's usual to keep only declarations in the header (to prevent the kind of compiler error that Dirk encountered) with the declaration living elsewhere in a source file.

inst/include/sampler.h
```diff
@@ -332,10 +431,12 @@ namespace smc {
...
 
-        //Normalise the weights to sensible values....
-        double dMaxWeight = arma::max(pPopulation.GetLogWeight());
-        pPopulation.SetLogWeight(pPopulation.GetLogWeight() - (dMaxWeight)*arma::ones(N));
+        //Estimate the normalising constant
+        dlogNCIt = CalcLogNC();
+        dlogNCPath += dlogNCIt;
 
+        //Normalise the weights
+        pPopulation.SetLogWeight(pPopulation.GetLogWeight()  - dlogNCIt*arma::ones(N));
```
adamjohansen: Do you actually need to create the vector of ones here (glancing at the Armadillo docs, isn't arma::vec - double well defined)? If you do need the extra vector, it might be sensible to have a static N-vector of ones available to avoid the overheads.

inst/include/population.h
LeahPrice: Great, thank you. I've created a new source file to contain that and possible other general use function definitions (it didn't fit naturally under the current source files).

inst/include/sampler.h
LeahPrice: Good point, thanks.

PR #8, opened 2017-07-28
Roxygen documentation

LeahPrice:
Adding Roxygen comments to automatically produce .Rd files.


LeahPrice:
I am unsure exactly where the file inst/sampleData/pf-data.csv came from. I can't find any mention of a special data source in the SMCTC paper and it makes sense that it would be simulated from the model for the example it's used in, so that's what I've assumed. Please correct me if I'm wrong.

eddelbuettel:
Please consider restoring the .csv file.  

It is somewhat rude to make changes to our code without even asking,   That is what issue tickets are for.

rcppsmc.Rproj
```diff
@@ -14,3 +14,4 @@ LaTeX: pdfLaTeX
 
 BuildType: Package
 PackageInstallArgs: --no-multiarch --with-keep.source
+PackageRoxygenize: rd,collate,namespace,vignette
```
eddelbuettel: I generally do not let my packages do anything but the rd part.

Please refrain from build changes without first establishing agreement.

ChangeLog
```diff
@@ -1,3 +1,29 @@
+2017-07-28  Leah South  <leah.south@hdr.qut.edu.au>
+
+    * DESCRIPTION: Setting LazyData to true.
+    * inst/sampleData/pf-data.csv: Removed.
+    * data/pfdata.rda: Replaces pf-data.csv.
```
eddelbuettel: I would prefer if you restored the csv file in the package.

If you think you must have an .rda, then keep the data file in, sat, `inst/rawdata`.

eddelbuettel:
It is nice to have roxygen but I am on the fence as to whether it is worthwhile rewriting existing documentation.

LeahPrice:
I'm sorry you found my changes to be rude, I certainly didn't mean for them to be. From previous discussions via email and from the plan on the issue ticket containing this item, I was under the impression that we were in agreement about these changes. The change to the data was originally made in a response to your comment in an R file. I don't have a strong preference and was happy with the previous documentation. We can remove this entire pull request if you like.

eddelbuettel:
You continue to blast waterfall-alike massive changes at us.  That is not a great process, irrespective of the fact that we have been doing it for a few days now.  

Let's try to find a better process. We can still merge this, but I want us to come to an understanding of a better working process.

LeahPrice:
Yes I agree we need to find a better working process as it seems I keep misinterpreting you. This sort of thing is so much easier in person! I thought the issue was that my changes were too small so I was trying to create a pull request that had one idea but wasn't too small. If it helps, there are a lot of changed lines here but the documentation for the existing things is all the same.

R/RcppSMC.R
```diff
@@ -0,0 +1,38 @@
+#' A package for general implementation of SMC algorithms.
+#'
+#' RcppSMC can be used to implement SMC algorithms ranging from
+#' simple particle filters to the SMC samplers of Del Moral, Doucet and Jasra (2006) 
+#' within a generic framework. For a general introduction to SMC methods, see
+#' Doucet, Freitas and Gordon (2001) and Doucet and Johansen (2008).
+#'
+#' RcppSMC is based around an Rcpp integration of the SMC template class library (Johansen, 2009).
+#'
+#' At present, three examples are included:
```
adamjohansen: Is this true any more? Don't you also have your tempered SMC sampler example?

It /is/ a pain keeping top level documentation up to date and we can certainly sanity check everything towards the end, but it always seems worth mentioning even trivial details when spotted.

R/bspfGaussianOptimal.R
```diff
@@ -1,8 +1,51 @@
...
+#' 
+#' @author		Adam M. Johansen, Dirk Eddelbuettel and Leah F. South
+#' @keywords	programming
+#' @export
 blockpfGaussianOpt <- function(data=c(), particles=1000, lag=5, plot=FALSE) {
 
-    if (length(data == 0)) {
-       #Include some error handling here
-       return;
+    if (length(data) == 0) {
```
adamjohansen: Given that this is just an illustrative example I guess that this is okay, but I'd shy away from doing this kind of thing in general.

Imagine calling a function within a script which is supposed to do something with your data, but omitting one argument in a long list (it happens to us all...); is it better for the code to fail or to run in an unexpected way producing a line of text in the output which could easily be missed? The answer, of course, depends on the context but I tend to err on the side of paranoia.

R/pfLineartBS.R
```diff
@@ -1,8 +1,55 @@
...
+#' res <- pfLineartBS(onlinePlot=pfLineartBSOnlinePlot)
+#' 
+#' @author		Adam M. Johansen, Dirk Eddelbuettel and Leah F. South
+#' @keywords	programming
+#' @export
+pfLineartBS<- function(dat, particles=1000, plot=FALSE, onlinePlot) {
 
     # if no data supplied, use default
-    if (missing(data)) data <- getPfLineartBSData()
+    if (missing(dat)){
```
adamjohansen: And now I see that the gung ho approach was entirely due to Dirk and I, but I think the comment remains valid. It's fine for these toy examples to make it easy for the user to see pretty pictures but we should avoid this with any code that might get used for anything important...

adamjohansen:
Sorry to come rather late to this...

It's great to see documentation appearing. I haven't made much use of Roxygen myself and don't have a position on it, anything which works to produce reasonable documentation is fine by me. This essentially seems fine to me.

It _would_ be good to settle into a pattern of commits which doesn't involve +522,-153 type changes because it's really hard for anyone to know what they've done in the longer term. It's also a good idea to separate structural changes from other things -- doing other things at the same time as what looks like a documentation update makes it really easy for changes to get lost.

Are we getting close to the stage where the work you've done is in the main repo? Once we reach that point it would definitely make sense to try to work with frequent smaller commits... 

Incidentally, if there are any things that you think should be done differently (you're very probably right) then we can discuss via whatever medium makes sense but as Dirk hinted it's generally a good idea to discuss (via appropriate mailing lists or fora for larger projects) proposed changes before sending code because (a) developers are people and (b) it's a good idea both to show that you've thought about things carefully and to check that there weren't good reasons for the way things used to me.

eddelbuettel:
>  It would be good to settle into a pattern of commits which doesn't involve +522,-153 type changes

I agree!  

Let's maybe hold this one PR back a bit, and try to focus on the other part of the task list.

LeahPrice:
Thanks Adam, I see your point about the error handling in the examples. I can change that in a future pull request but to keep things neat I won't do it along with the linear regression example pull request.

Unfortunately (or fortunately, depending on how you look at it), there are still a few more changes that are finished but need to be pulled across. [Issue ticket 6](https://github.com/eddelbuettel/rcppsmc/issues/6) gives more details on this.

I was about to do a pull request for the linear regression example after seeing you were okay with it, but it involves +730,-0 changes so perhaps that's too much. There are a couple of things I'd like to get your opinion on about this actually, but I'll take that discussion to [the relevant issue ticket](https://github.com/eddelbuettel/rcppsmc/issues/9).

eddelbuettel:
Well +730,-0 may be way easier as it is all new code.  

But simultaneous changes across multiple files/issues/topics smushed into a single PR is bad.

adamjohansen:
True -- and we don't want to obsess over numbers, sometimes big commits are the right answer. (And I was being a little unfair previously as 500 lines of comments aren't really the same as 500 lines of code).

eddelbuettel:
I'm closing this one.  We can revisit the need for conversion of existing documentation at a later time.

PR #15, opened 2017-08-04
Updating news and fixing data note

LeahPrice:
Silencing --as-cran R CMD check note about data and updating the NEWS file. I can add more detail to NEWS if this isn't detailed enough.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2017-08-04  Leah South  <leah.south@hdr.qut.edu.au>
+
+	* R/LinReg.R: Silencing R CMD check --as-cran by not loading
+    data(radiata) into the global workspace.
+	* R/LinRegLA.R: Idem.
+
 2017-08-03  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@
 	* R/LinReg.R: Silencing R CMD check --as-cran by not loading
     data(radiata) into the global workspace.
 	* R/LinRegLA.R: Idem.
+	* inst/NEWS: Adding a line for each of the recent major changes.
 
 2017-08-03  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2017-08-04    <adam.johansen@gmail.com>
+
+	* R/LinRegLA.R: Suppressing redundant call to data().
+	* R/LinReg.R: Idem.
+
 2017-08-04  Leah South  <leah.south@hdr.qut.edu.au>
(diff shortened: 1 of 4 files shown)


eddelbuettel:
Looks good at a first glance.  Common idiom is to keep each entry to a line, max two.  We can polish in subsequent ones as we aren't done overall given all your work.

But bedtime here now... or soon.

LeahPrice:
Thanks Dirk. Yeah I thought it must be pretty late there. Good night!

adamjohansen:
That all looks good to me; if we keep NEWS updated from here on and polish towards the end of the summer then we should be in good shape, I think.

adamjohansen:
That all looks good to me; if we keep NEWS updated from here on and polish towards the end of the summer then we should be in good shape, I think.

R/LinReg.R
```diff
@@ -1,6 +1,6 @@
 LinReg<- function(model, particles=1000, plot=FALSE) {
 
-    data(radiata)
+    data(radiata, envir = environment())
```
eddelbuettel: Is this really the best we can do / should do?  I haven't looked at this (and the manual) in some time but I believe I have dealt with this differently in the past.

adamjohansen: I have to take some responsibility for this (https://github.com/eddelbuettel/rcppsmc/pull/13#issuecomment-319898169); the writing R extensions docs (https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Data-in-packages) weren't very prescriptive about The Right Way to deal with this and I have no strong views.

eddelbuettel: We can try. I sort-of rediscover / relearn this every time.  All I know right now is that I stick it into the `data/` directory and things but it has been a while since I added data to a package...  Just checked one of the recent ones and yes, I do use `LazyData: true` there and need no `environment()`.  Maybe that is just it.

adamjohansen: I've tried. This passes all of the sanity checks I tried with R 3.4.1 and Travis seems happy with it.

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2017-08-04    <adam.johansen@gmail.com>
```
eddelbuettel: The format is date, two spaces, name, two spaces, email.

That said, it is also uncommon to write these for someone else.

eddelbuettel:
Nice.  Passes fine with R-devel too.

ChangeLog
adamjohansen: Oops, sorry. Missing user-full-name specification in a .emacs file which I'd overlooked. C-x 4 a was indeed what I used, but apparently on a misconfigured system and without paying enough attention on a Friday evening.

eddelbuettel: Not to worry. I _finally_ started a repo dotfiles for just those types of things across machines.

LeahPrice:
Thanks Adam!

PR #19, opened 2017-08-06
Parameter object and adaptation

LeahPrice:
This commit is the start of making it easier to perform adaptation in SMC.

The main changes here are:
- Adding a template parameter for the algorithm parameters to the sampler object
- Creating a base class for adaptation
- Doing the MCMC repeats within the library

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,24 @@
-2017-08-04    <adam.johansen@gmail.com>
+2017-08-06  Leah South  <leah.south@hdr.qut.edu.au>
+
+	* inst/include/algParam.h: A base class to contain
+	additional algorithm parameters and virtual functions to
+	adapt them.
+	* inst/include/moveset.h: Changing the MCMC step to perform
+	multiple iterations within the library rather than on user end.
+	* inst/include/sampler.h: Adding an additional template
+	parameter for additional parameters and allowing for
+	optional adaptation at multiple points.
+
+	* src/LinReg.cpp: Boolean return for MCMC function and
+	setting MCMC repeats separately.
+	* src/LinReg.cpp: Idem.
+	* inst/include/LinReg.h: Idem.
+	* inst/include/LinReg_LA.h: Idem.
+
+	* src/RcppExports.cpp: Regenerated.
+	* R/RcppExports.R: Idem.
+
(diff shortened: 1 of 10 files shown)


LeahPrice:
I changed the MCMC step code a little so that the loop over the repeats is done within the library and users write code for a single MCMC step for a single particle. This seemed helpful for adapting the number of MCMC repeats (it is easier to work out the acceptance probability from previous steps if you number the number of repeats).

I think in a future pull request I’ll store more of the calculations for data annealing SMC. I’m essentially doubling up on the likelihood and prior calculations because I’m recalculating them for the current particle at every iteration.

R/RcppExports.R
```diff
@@ -2,26 +2,26 @@
 # Generator token: 10BE3573-1514-4C36-9D1C-5A225CD40393
 
 blockpfGaussianOpt_impl <- function(data, part, lag) {
-    .Call(`_RcppSMC_blockpfGaussianOpt_impl`, data, part, lag)
+    .Call('RcppSMC_blockpfGaussianOpt_impl', PACKAGE = 'RcppSMC', data, part, lag)
```
eddelbuettel: You are using an Rcpp version older than 0.12.12.  

Please upgrade, and re-create the file, or explain why you must use on old version, possibly over at the Rcpp repo with an issue ticket there.

ChangeLog
```diff
@@ -1,4 +1,24 @@
...
+	* src/LinReg.cpp: Boolean return for MCMC function and
+	setting MCMC repeats separately.
+	* src/LinReg.cpp: Idem.
+	* inst/include/LinReg.h: Idem.
+	* inst/include/LinReg_LA.h: Idem.
+
+	* src/RcppExports.cpp: Regenerated.
+	* R/RcppExports.R: Idem.
+
+2017-08-04  Adam M. Johansen  <adam.johansen@gmail.com>
```
eddelbuettel: Thanks for cleaning that up.

R/RcppExports.R
LeahPrice: Thanks for picking up on that. I've fixed it up now.

eddelbuettel: Thanks.

adamjohansen:
Thanks, both -- looks good.

PR #20, opened 2017-08-07
Commonly used adaptation methods

LeahPrice:
Adding a class containing commonly used parameters and methods to adapt them (for static Bayesian models).

### DIFF
--- a/src/staticModelAdapt.cpp
+++ b/src/staticModelAdapt.cpp
@@ -1,0 +1,128 @@
+// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
+//
+// staticModelAdapt.cpp: A class containing parameters and functions to update
+// these parameters in the context of static Bayesian models. The methods to
+// estimate the empirical covariance and its Cholesky decomposition are applicable
+// for applications where the particle values are a vector of doubles. The methods
+// to compute the next 'temperature' are relevant to likelihood annealing SMC where the
+// power posteriors are defined by P_t(theta|y) \propto P(y|theta)^\gamma_t P(theta).
+// Here y is observed data, theta denotes the parameters of the model and gamma
+// denotes the 'temperatures' which start at 0 (the prior) and finish at 1
+// (the posterior).
+//
+// Copyright (C) 2017         Dirk Eddelbuettel, Adam Johansen and Leah South
+//
+// This file is part of RcppSMC.
+//
+// RcppSMC is free software: you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// RcppSMC is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


src/staticModelAdapt.cpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /// Chooses the next temperature such that a desired conditional ESS is maintained.
+    ///
+    /// \param logweight An armadillo vector containing the logarithm of the current particle weights.
+    /// \param loglike An armadillo vector containing the log likelihood of the current particle values.
+    /// \param desiredCESS The target conditional ESS for the next temperature (generally fixed).
+    void staticModelAdapt::ChooseTemp(const arma::vec & logweight, const arma::vec & loglike, double desiredCESS) {
+        double temp_curr = temp.back();
+        if (CESSdiff(logweight,loglike,1-temp_curr,desiredCESS)>=0){
```
adamjohansen: Might it make sense for the comparison to be with -0.01 (or the tolerance) to be consistent with the inner workings of the bisection function?

src/staticModelAdapt.cpp
```diff
@@ -0,0 +1,128 @@
...
+        double f_a = CESSdiff(logweight,loglike,a-curr,desiredCESS);
+        double f_b = CESSdiff(logweight,loglike,b-curr,desiredCESS);
+        if (f_a*f_b>0){
+            Rcpp::stop("Bisection method to choose the next temperature failed");
+        } else{
+            double m, f_m, err;
+            m = (a+b)/2.0;
+            f_m = CESSdiff(logweight,loglike,m-curr,desiredCESS);
+            err = 10;
+            while (err > 0.01){
```
adamjohansen: It might be good if it were possible to choose this tolerance without changing a hard-coded constant. In some applications the desired CESS might be larger than 0.99 and a tolerance of 0.01 would be problematic in those settings.

src/staticModelAdapt.cpp
```diff
@@ -0,0 +1,128 @@
...
+        cholCov = arma::chol(empCov);
+    }
+
+    /// Calculates the number of MCMC repeats based on the results of the most recent set of MCMC moves.
+    ///
+    /// \param acceptProb The proportion of accepted MCMC steps in the most recent iteration.
+    /// \param desiredAcceptProb The desired probability of a successful move for each particle.
+    /// \param initialN The initial number of MCMC repeats.
+    /// \param maxReps The maximum number of MCMC repeats.
+    int staticModelAdapt::calcMcmcRepeats(double acceptProb, double desiredAcceptProb, int initialN, int maxReps){
```
adamjohansen: It's a bad idea to compare doubles with integers using ==, usually (the finite precision issue), might it make sense to replace these comparisons with 
abs(acceptProb - 1) <= 10**-9
or similar?

src/staticModelAdapt.cpp
LeahPrice: Thanks Adam for this and for all of your other good points. I should have picked up on at least some of these myself -- sorry about that. I've changed this and a couple of other minor things (like writing 1.0 instead of 1 if it's a double) and will do another commit when the check finishes.

adamjohansen:
Great, thanks.

PR #21, opened 2017-08-08
MCMC Repeats

LeahPrice:
I accidentally left off an absolute value in one of the calculations for the MCMC repeats. This pull request is focused on
- fixing that issue
- storing the MCMC repeats in the history
- Adding a couple of items to NEWS (about the pull requests so far for adaptation)

Sorry this is a bit of a step backwards in terms of useful pull requests, but I figured it was preferable to do this separately to the next pull request which contains the examples. I have the code for that available [here](https://github.com/LeahPrice/rcppsmc/commit/2013e83e05f8da464146e1dbc4a0817d7ce69b5e) so I will create a pull request for it after this one is sorted.

I had issues with Gitkraken today and ended up having to set up the clone on my desktop again. I don't _think_ that caused any problems.

Can I please get your opinion on the Travis CI note below?

>   installed size is  6.8Mb
>   sub-directories of 1Mb or more:
>     libs   6.5Mb

### DIFF
--- a/src/staticModelAdapt.cpp
+++ b/src/staticModelAdapt.cpp
@@ -118,7 +118,7 @@ namespace smc {
     int staticModelAdapt::calcMcmcRepeats(double acceptProb, double desiredAcceptProb, int initialN, int maxReps){
         if (acceptProb + 1.0 <= 1e-9){
             return initialN;
-        } else if (acceptProb - 1.0 <= 1e-9){
+        } else if (fabs(acceptProb - 1.0) <= 1e-9){
             return 1;
         } else if (acceptProb <= 1e-9){
             return maxReps;
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,10 @@
 2017-08-08  Leah South  <leah.south@hdr.qut.edu.au>
 
-	* src/staticModelAdapt.cpp: Adding missing absolute value
-	to calculations of MCMC repeats.
+	* src/staticModelAdapt.cpp: Adding missing absolute value 
+	to calculations of MCMC repeats. 
+
+	* inst/include/history.h: Adding MCMC repeats to history.
+	* inst/include/sampler.h: Idem.
 
 2017-08-07  Leah South  <leah.south@hdr.qut.edu.au>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2017-08-08  Leah South  <leah.south@hdr.qut.edu.au>
+
+	* src/staticModelAdapt.cpp: Adding missing absolute value
+	to calculations of MCMC repeats.
+
 2017-08-07  Leah South  <leah.south@hdr.qut.edu.au>
 
 	* inst/include/staticModelAdapt.h: A class for commonly used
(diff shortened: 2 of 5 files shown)


src/staticModelAdapt.cpp
```diff
@@ -118,7 +118,7 @@ namespace smc {
     int staticModelAdapt::calcMcmcRepeats(double acceptProb, double desiredAcceptProb, int initialN, int maxReps){
         if (acceptProb + 1.0 <= 1e-9){
             return initialN;
-        } else if (acceptProb - 1.0 <= 1e-9){
+        } else if (fabs(acceptProb - 1.0) <= 1e-9){
```
adamjohansen: I'm not sure what @eddelbuettel thinks about abs/fabs but I'd be inclined to just use std::abs in C++ code. Not that it is of any consequence... actually,on this instance I would have been inclined to say 
if ((acceptProb- 1.0) >= -1e-9) { ...
as there's not much danger of it being larger than 1+1e-9.

adamjohansen:
On the size of the library directory, I'd been wondering about the root cause of this and haven't got around to looking yet. The RcppSMC.so file is rather larger considering what it does -- is it possible we're statically linking against something which we shouldn't be?

If it's the size the library needs to be then that's fair enough, but if we're doing something daft that's inflating it to a size rather larger than that of RcppArmadillo.so (which feels likely under the circumstances) then we should certainly address that.

I'll have an investigate later on, but I'd better get on with the day job now.

eddelbuettel:
I also prefer `std::abs()` from C++, but have of course use `fabs()` in the past.

The "size of the library directory" is a C++ side effect we cannot do anything about, sadly.   I get that with many packages.

adamjohansen:
Thanks, Dirk, if it's unavoidable I'll stop worrying about it...

eddelbuettel:
And friend has played some tricks to get `strip` to act on it after the build.  That, and not using `-g`, can bring the size down.  But both are somewhat non-standard.

src/staticModelAdapt.cpp
LeahPrice: Thanks Adam. I'll use that suggestion.

LeahPrice:
Thanks Adam and Dirk. I won't worry about the size note too much for now then.

eddelbuettel:
Hi Leah, I see these two uses of `abs()` and `fabs()`:

```bash
~/git/rcppsmc(master)$ ag "abs\(" src/staticModelAdapt.cpp
68:                err = abs(f_m);
121:        } else if (fabs(acceptProb - 1.0) <= 1e-9){
~/git/rcppsmc(master)$ 
```

If it were I'd be explicit and use `std::abs` in both cases.  Together with `cmath.h` this should give you `double abs(doube)` from the `std` namespace.   If the compiler does not complain we _probably_ already use it on the first call.

LeahPrice:
Hi Dirk, for some reason the compiler didn't complain about it but I checked and it wasn't using the double version. Thanks for picking up on that. I'll change the first call to std::abs and the second to Adam's suggestion in the next pull request.

PR #22, opened 2017-08-08
Adaptation in the Linear Regression Example

LeahPrice:
Adapting the temperature schedule, number of MCMC repeats and the random walk covariance matrices in the likelihood annealing SMC implementation of the simple linear regression example.

Also correcting a couple of issues with calls to abs() from a previous pull request.


R/LinRegLA_adapt.R
```diff
@@ -0,0 +1,17 @@
...
+    } else{
+        stop("Please choose a valid model (1 or 2).")
+    }
+    
+    res <- LinRegLA_adapt_impl(as.matrix(Data), particles, resampTol, tempTol)
+
+    invisible(res)
+}
+
+## silence a NOTE from 'R CMD check --as-cran'
```
adamjohansen: Further to earlier discussion about this....

If I'm understanding a somewhat aged discussion correctly, BDR seems to suggest that we should just use the fully-qualified name to refer to the data set (http://r.789695.n4.nabble.com/no-visible-binding-for-global-variable-for-data-sets-in-a-package-tt4696053.html#a4696100) and use RcppSMC::radiata rather than radiata.

If I understood correctly, that obviates the need for a workaround like this. 

Although my instinct is to do as BDR directs unless there's a good reason to do otherwise, as with all such things I bow to @eddelbuettel and his vastly more extensive knowledge, so it's quite possible that I've misunderstood entirely or that there is some other reason for explicitly dealing with the checker in this way.

src/LinReg_LA_adapt.cpp
```diff
@@ -0,0 +1,178 @@
...
+        
+        double MH_ratio = exp(myParams->GetParams().GetTemp(lTime)*(value_prop.loglike - value.loglike) + value_prop.logprior - value.logprior);
+        
+        if (MH_ratio>R::runif(0,1)){
+            value = value_prop;
+            return TRUE;
+        }
+        return FALSE;
+    }
+}
```
adamjohansen: Missing newline...

R/LinRegLA_adapt.R
eddelbuettel: I don't use the `data()` features often and recall fighting it, so as opinionated as I regularly am, I will refrain from piping in here.  At least one recent-ish package also uses `LazyData: true` as we do here.

adamjohansen: It's not something that I'm really concerned about, beyond a vague instinct to avoid code which is there only to suppress warnings.

eddelbuettel: Oh yes I am all ears for best use of `data()`.

LeahPrice: Great, thanks for the suggestion. It worked well so I just committed the changes.

src/LinReg_LA_adapt.cpp
LeahPrice: Thanks. I've fixed this up now.

PR #28, opened 2017-11-14
Add copy constructor and assignment overloading

### LINKED ISSUE #25: Longer-term TODO List
In addition to adding the linear Gaussian example, I have been thinking of doing some of the following minor changes. These should all be relatively quick, besides the ones based on generalizing the examples. Could I please get your thoughts on these @eddelbuettel and @adamjohansen? Feel free to edit this post.

- [ ] Add an example of << ostream operator overloading
- [ ] Allow for proper deep copies of pAdapt and adaptBelong
- [ ] Move the additional parameters for all proposals into parameter objects (?)
- [ ] Add verbose/quiet option for PMMH and output information at every 100 iterations if verbose
- [ ] Add the additional parameters to the history object
- [ ] Generalize the linear regression example
- [ ] Generalize other examples (?)
- [ ] Go back to Hungarian notation (or at least ensure it is consistent)
- [ ] Do we still want to have R wrapper functions for basically every C++ function?
- [ ] Do we still want the folder src/compareGSLtoR in its current location?

I also need to include the updated NEWS and TODO files in the next PR.

binxiangni:
Learning some computational statistics methods this semester and trying to apply my knowledge to fix #25.  One question: how do you do unit tests in this package?

### DIFF
--- a/inst/include/sampler.h
+++ b/inst/include/sampler.h
@@ -1,10 +1,10 @@
 // -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
 //
-// sampler.h: Rcpp integration of SMC library -- sampler object 
+// sampler.h: Rcpp integration of SMC library -- sampler object
 //
 // Copyright (C) 2008 - 2009  Adam Johansen
 // Copyright (C) 2017         Adam Johansen, Dirk Eddelbuettel and Leah South
-//  
+//
 // This file is part of RcppSMC.
 //
 // RcppSMC is free software: you can redistribute it and/or modify it
@@ -43,30 +43,30 @@
 ///Specifiers for various resampling algorithms:
 namespace ResampleType
 {
-    enum Enum { MULTINOMIAL = 0, 
-        RESIDUAL, 
-        STRATIFIED, 
+    enum Enum { MULTINOMIAL = 0,
+        RESIDUAL,
+        STRATIFIED,
         SYSTEMATIC };
 }
 
 ///Specifiers for various path sampling methods:
 namespace PathSamplingType
 {
-    enum Enum { TRAPEZOID2 = 0, 
-        TRAPEZOID1, 
+    enum Enum { TRAPEZOID2 = 0,
+        TRAPEZOID1,
         RECTANGLE};
 }
 
 
-///Storage types for the history of the particle system. 
+///Storage types for the history of the particle system.
 namespace HistoryType
 {
-    enum Enum { NONE = 0, 
+    enum Enum { NONE = 0,
         RAM};
 }
 
 namespace smc {
-    
+
     /// An empty class for use when additional algorithm parameters are not required.
     class nullParams{};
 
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Sorry, I am lost:  where does #25 say anything about a copy constructor>

In general, I am a fan of first opening issue tickets, discussing what you want to do and getting consensus for it to get done.  Throwing "random" pull request over the fence and asking us to review is a little risky as everybody is busy.

binxiangni:
Sorry, I'll open a new issue next time to contribute. The fourth item of the checklist in #25 is `Add sampler copy constructor`.

eddelbuettel:
Got it. Overlooked this. 

This project is now a bit of a continuation of Leah's GSoC.  If you have spare cycles, I may have something else (or it may be too small...).

binxiangni:
I have a long Thanksgiving break and trying to find something to work on.

LeahPrice:
This looks good to me, thanks Binxiang! It would be good to have you involved if you'd like to be. I'd also be happy to chat about computational statistics if you'd like. My PhD is in comp stats.

inst/include/sampler.h
```diff
@@ -246,6 +247,75 @@ namespace smc {
...
+        ///Structure used internally for resampling.
+        uRSCount = sFrom.uRSCount;
+        ///Structure used internally for resampling.
+        uRSIndices = sFrom.uRSIndices;
+
+        ///The particles within the system.
+        pPopulation = sFrom.pPopulation;
+        ///The set of moves available.
+        Moves = sFrom.Moves;
+        /// The additional algorithm parameters.
```
adamjohansen: Sorry for being rather late with this, I've been swamped with other things.

Thanks for this, Binxinag, it's always good to sort out these small issues. 

However, there is a bit of a complication, at this stage the code copies a pointer and also an indicator of whether the object pointed to belongs to a sampler. If you copy a sampler which owns pAdapt and delete both of them then they will both delete the same object.

I suspect that the cleanest option would be to do a deep copy of the object pointed to and then to set adaptBelong to TRUE regardless of what the original configuration was, but I haven't thought very much about this.

Any other thoughts on this?

binxiangni: Got it. Perhaps we could do it like 
```
 *pAdapt = *sFrom.pAdapt;
 adaptBelong = sFrom.adaptBelong;
```
Then even if `adaptBelong == true`, we do not delete the same object.

binxiangni: Then one new problem may come out. When pAdapt is null at first, *pAdapt may bring segment fault. :(

adamjohansen: Quite so. This is where pAdapt would be initialised so to go down that route you'd need to allocate memory before assigning to it -- but that wouldn't need much more than an additional 
`pAdapt = new...`
assuming that the deep copy model is the approach we want to go with here.

Also, if we do do this then it should probably be 
`adaptBelong = true`
rather than 
`adaptBelong = sFrom.adaptBelong` 
because even if the source object doesn't own the object pointed to by its `pAdapt` the copied object does.

LeahPrice: Thanks Adam for picking up on this. I'd forgotten about this issue of dealing with the pointer in the copy. I think I put this off because I didn't have a great solution at the time but I like the sound of your solution.

adamjohansen: There's been a history of omitting the copy constructor because it was fiddly for various reasons dating right back to the first implementation. I think we're closing in on something that will at least work -- it will be great to have something in place, but we should remember to be clear in any documentation exactly what the behaviour on copying is.

(The only obvious alternative that I can see would be to always copy the pointer and then set `adaptBelong = false` but that runs into problems if the original sampler owned the object and is destroyed before the copy... but if anyone has other thoughts they'd be very welcome.)

PR #32, opened 2018-01-31
safe smc::sampler copy constructor

### LINKED ISSUE #31: Some Self-contained Improvements
This is a list of some relatively straightforward and self contained tasks which it would be good to get done. 

- [x] Make the smc::sampler copy constructor safe.
- [ ] Improve history getters (so that users can directly call things like `Sampler.GetMcmcRepeats(time)` and `Sampler.GetPopulation(time)`)
- [ ] Add an example of  `<<` ostream` operator overloading

In more detail:

- **Make the smc::sampler copy constructor safe.** In #28 a copy constructor was added, filling a long-standing gap in the code. However, there remains a memory allocation issue which is described in the discussion of #28. The simplest way to fix this would be to deep copy the object which contains adaptation information and to ensure that the new copy of the sampler always owns the copied object.
- **Keep track of more log likelihood and log prior calculations in the data annealing linear regression example (to avoid unnecessary extra calculations)** In this example it would be possible to cache more of the likelihood and prior density evaluations to avoid redundant re-evaluations. This should improve execution times.
- **Improve history getters** The `smc::history` class contains information about the previous generations of the sampler in samplers for which this information is being stored. At present accessing thijs information is rather cumbersome, it would be nice to add to the `smc::sampler` object some helper functions to return the quantities stored within this history object directly to make the end user experience better.
- **Add an example of  `<<` ostream operator overloading** At present the stream output operator is overloaded for `smc::sampler` objects but this code hasn't been used in some time. It would be good to check that this code still functions as expected and then to add an example of its use to one of the example samplers.

mlysy:
Dear RcppSMC Developers,

I think this resolves #31 point 1.


eddelbuettel:
You're a star.  But we are collecting some of the ideas on #31 to have possible tests for students in [this year's Google Summer of Code](https://summerofcode.withgoogle.com/) where we [posted another proposal](https://github.com/rstats-gsoc/gsoc2018/wiki/RcppSMC).  By all means please feel free to chip in if you have additonal ideas, particularly for the earlier "test stage" (ie not too trivial but not to insurmountable) as well as for possible three-month focus ideas.

We'll likely merge the PR but I am a little fried now; we'll see what Adam says in a few...

inst/include/sampler.h
```diff
@@ -273,10 +274,21 @@ namespace smc {
...
         algParams = sFrom.algParams;
-        /// An object for adapting additional algorithm parameters
-        pAdapt = sFrom.pAdapt;
-        ///A flag to track whether the adaptation object needs to be included in this destructor.
-        adaptBelong = sFrom.adaptBelong;
+        if(sFrom.adaptBelong) {
+            // this can only happen if the default adaptMethods was used,
+            // i.e., no call to SetAdaptMethods
+            pAdapt = new adaptMethods<Space,Params>;
+            adaptBelong = 1;
```
adamjohansen: Don't we also need a
`*pAdapt = *sFrom.pAdapt`
here?

I'd expect a copy of a sampler to end up in exactly the same configuration as the one that was being copied.

inst/include/sampler.h
```diff
@@ -273,10 +274,21 @@ namespace smc {
...
+        } else {
+            // this can only happen if SetAdaptMethods was called,
+            // i.e., pAdapt points to an external object which should not be deleted with sampler
+            pAdapt = sFrom.pAdapt;
+            adaptBelong = 0;
```
adamjohansen: I'm not sure this is the right behaviour.

If we just copy a pointer like this, then we end up in a situation in which iterating either the original sampler or the new copy changes the configuration  of _both_ samplers which is clearly tricky.

It might not be completely trivial to resolve this in an elegant way; there was a reason that this constructor was absent prior to #28 and that was that it's a little delicate to do and not quite clear what the correct behaviour is. Perhaps this pointer to an object model wasn't such a good idea.

inst/include/sampler.h
LeahPrice: I would have thought that this one would be okay. If adaptBelong is 1 then SetAdaptMethods hasn't been called at any point so pAdapt is empty. I *think* the above will produce something that's equivalent but independent.

inst/include/sampler.h
LeahPrice: Good point. I find this stuff quite difficult and I'm not sure how to handle it without either having slicing or dependence.

[This SO discussion](https://stackoverflow.com/questions/12255546/c-deep-copying-a-base-class-pointer) suggests doing a virtual function for the 'clone', but that seems annoying for users.

The suggestions in Bjarne Stroustrup's book seem to be centred around using clone functions too. The options he suggests are along the lines of preventing copying or "make the copy operations private or protected, and provide a clone() function to allow a user to request a non-slicing copy." I.e. using clone as above.

inst/include/sampler.h
adamjohansen: That might be true with the current implementation, but there's a certain lack of robustness in assuming you can trace the execution path to the current configuration. Any future changes could break things in a way that would be rather subtle.

In particular, I think it's going to be difficult to address the other issue without introducing a mechanism by which a copied sampler can end up owning pAdapt when the object being copied did not.

As `adaptBelong` only indicates memory ownership I'd be inclined to include an explicit copy even if it may be redundant at this instant.

inst/include/sampler.h
adamjohansen: Well, if we want to allow duplication of derived types from a base class I can't see a way of avoiding a virtual mechanism for doing it so unless there's some cunning way of doing this I'm not sure it can be avoided.

Perhaps it would make sense to agree that this PR at least addresses the memory-safeness (perhaps moduluo the explicit copy I suggested above) and add another item to the list to deal with allowing for proper deep-copies of these objects?

LeahPrice: Okay that seems like a good plan to me as well.

mlysy: Dear All, thanks for catching this.

Actually I was wondering why `SetAlgParam` passes by value whereas `SetAdaptMethods` does not, as the former does bypass this copying problem...

adamjohansen:
Any objection to us merging this, @eddelbuettel ? I think it's progress but does still leave an open issue of how best to deal with the complications arising from inheritance that I think is probably too complicate for the "easy to address" problems list on reflection.

Also, from a design philosophy point of view, are you happy with me adding a (currently redundant) copy for future robustness as discussed above?

eddelbuettel:
Ahh, I may have I misread your comments as more critical than you meant.  Yes, incremental is good.  Let's merge now and you can tune it as time permits.  All good :)

adamjohansen:
I have a tendency to sound more negative than I intend, my fault... there's still an issue with unintended consequences of copying certain types of object, but that's going to be a pain.

Issue #34, opened 2018-03-02
Bug in pfNonlinBS?


mdbauer:
There appears to be a bug in line 73 of `pfnonlinbs.cpp`. For calculating the standard deviation of the filtering distribution, the auxiliary information passed to the integrand function should be the mean, not the standard deviation. That is, the second `resSD(n)` in 

``resSD(n) = sqrt(Sampler.Integrate(integrand_var_x, (void*)&resSD(n))); ``

should be replaced by `resMean(n)`.

(If I may also note a tiny inconsistency between the default parameters of `simNonlin()` and the parameters assumed in `pfNonlinBS()`: for the variance of the initial observation, the simulation uses 10 while the bootstrap filter assumes 4. But this discrepancy is immaterial.)

Thank you for all the amazing work you guys contribute to statistics, finance, and #rstats. I'm a huge fan!

adamjohansen:
Quite so! Thank you for pointing that out... embarrassing but at least easily fixed. I'll get to it over the weekend with any luck.

I can't recall any good reason for the inconsistency that you mention, so perhaps we should tidy that up at the same time.

eddelbuettel:
Because we overlooked the `closes #34` tag.  Thanks for noticing!

mdbauer:
I should note that with this bugfix the figure on [Dirk's RcppSMC page](http://dirk.eddelbuettel.com/code/rcpp.smc.html) looks very different. The current, incorrect figure has confidence intervals that are too wide and always include zero. The new, corrected figure would look something like this:
![nonlinbsfilter](https://user-images.githubusercontent.com/11216378/37444614-d79d845c-27cf-11e8-8c6c-45807749b620.png)

eddelbuettel:
Thanks for catching that too.  I replaced the underlying png file with a new one.  Should be better now.

PR #36, opened 2018-03-03
Bugfix

adamjohansen:
A fix to address #34 and to make the simulation code and particle filter use the same default value for the initial state variance.

Nothing controversial is there (mainly aimed at @eddelbuettel and @LeahPrice)?


LeahPrice:
This all looks good to me. Thanks Adam.

PR #38, opened 2018-03-15
PMMH verbouse and quite output every 100 iterations

tibi77:
I do not know if the data that I am printing is the one that you want so please take a look. @LeahPrice  @eddelbuettel  @adamjohansen

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -17,8 +17,8 @@ LinRegLA_impl <- function(Data, intemps, lNumber) {
     .Call(`_RcppSMC_LinRegLA_impl`, Data, intemps, lNumber)
 }
 
-nonLinPMMH_impl <- function(data, lNumber, lMCMCits) {
-    .Call(`_RcppSMC_nonLinPMMH_impl`, data, lNumber, lMCMCits)
+nonLinPMMH_impl <- function(data, lNumber, lMCMCits, verbouse) {
+    .Call(`_RcppSMC_nonLinPMMH_impl`, data, lNumber, lMCMCits, verbouse)
 }
 
 pfLineartBS_impl <- function(data, part, usef, fun) {
--- a/src/nonLinPMMH.cpp
+++ b/src/nonLinPMMH.cpp
@@ -34,7 +34,7 @@ using namespace nonLinPMMH;
 
 // nonLinPMMH_impl() function callable from R via Rcpp::
 // [[Rcpp::export]]
-Rcpp::DataFrame nonLinPMMH_impl(arma::vec data, unsigned long lNumber, unsigned long lMCMCits) {
+Rcpp::DataFrame nonLinPMMH_impl(arma::vec data, unsigned long lNumber, unsigned long lMCMCits, bool verbouse) {
 
     try {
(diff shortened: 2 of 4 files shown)


src/nonLinPMMH.cpp
```diff
@@ -97,6 +99,14 @@ Rcpp::DataFrame nonLinPMMH_impl(arma::vec data, unsigned long lNumber, unsigned
                 loglike(i) = loglike(i-1);
                 logprior(i) = logprior(i-1);
             }
+            if (i % 100 == 0 && verbouse){
+		Rcpp::Rcout << "========================="<< endl;
```
adamjohansen: This output looks a bit longer and more cluttered than is really needed, wouldn't:
`"Iteration: " << i << " Average acceptance probability: " << MH_ratio_total / i << ...`
suffice without taking half a dozen lines to output three or four numbers?

src/nonLinPMMH.cpp
```diff
@@ -97,6 +99,14 @@ Rcpp::DataFrame nonLinPMMH_impl(arma::vec data, unsigned long lNumber, unsigned
...
+		Rcpp::Rcout << "Iteration number: " << i << endl;
+		Rcpp::Rcout << "The MH_ratio is: " << std :: min(1.0, MH_ratio) << endl;
```
adamjohansen: This would be better described as "acceptance probability" than "MH_ratio" both for accuracy and interpretability. Similarly the next item is the average acceptance probability.

Personally, I'd only bother outputting the average, but that might be a matter of taste.

src/nonLinPMMH.cpp
```diff
@@ -97,6 +99,14 @@ Rcpp::DataFrame nonLinPMMH_impl(arma::vec data, unsigned long lNumber, unsigned
...
+		Rcpp::Rcout << "The MH_ratio_average is: " << MH_ratio_total / i << endl;
+		Rcpp::Rcout << "Logarithmic probability: " << loglike_prop << endl;
```
adamjohansen: I'm not sure you're outputting the right thing here; you'd ideally have the logarithm of the target density, not the logarithm of the likelihood of the most recently proposed value. That would amount to loglike(i) + logprior(i) and coul be described as _log posterior density_ (which it is up to normalization).

R/RcppExports.R
```diff
@@ -17,8 +17,8 @@ LinRegLA_impl <- function(Data, intemps, lNumber) {
     .Call(`_RcppSMC_LinRegLA_impl`, Data, intemps, lNumber)
 }
 
-nonLinPMMH_impl <- function(data, lNumber, lMCMCits) {
-    .Call(`_RcppSMC_nonLinPMMH_impl`, data, lNumber, lMCMCits)
+nonLinPMMH_impl <- function(data, lNumber, lMCMCits, verbouse) {
+    .Call(`_RcppSMC_nonLinPMMH_impl`, data, lNumber, lMCMCits, verbouse)
```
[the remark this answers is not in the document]
tibi77: We can make a linter of some kind to check for code format and so on.

PR #42, opened 2018-03-17
use std::pow instead of pow to fix solaris build

eddelbuettel:
We seem to fail to build on Slowlaris as seen [here](https://www.r-project.org/nosvn/R.check/r-patched-solaris-x86/RcppSMC-00install.html).

This may help.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,13 @@
 2018-03-17  Dirk Eddelbuettel  <edd@debian.org>
 
+	* inst/include/history.h: Use std::pow
+	* inst/include/sampler.h: Idem
+	* src/LinReg.cpp: Idem
+	* src/LinReg_LA_adapt.cpp: Idem
+	* src/nonLinPMMH.cpp: Idem
+	* src/pflineart.cpp: Idem
+	* src/pfnonlinbs.cpp: Idem
+
 	* DESCRIPTION (Version, Date): New minor version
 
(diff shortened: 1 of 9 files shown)


adamjohansen:
Thanks Dirk, this looks pretty uncontroversial to me... fingers crossed that it resolves the ambiguity.

eddelbuettel:
I tried it at r-hub in order to use solaris, but the builder had a network issue.  Log [is here](https://builder.r-hub.io/status/RcppSMC_0.2.0.3.tar.gz-18c33abf1b38437db1bee2febf8d7782).

eddelbuettel:
Wow. @gaborcsardi already fixed the [rhub issue I filed](https://github.com/r-hub/rhub/issues/126#event-1526998257) and I just checked the package with this PR there.  No issue.  (Now, I am not fully sure that it is the identical setup but probably close enough.

eddelbuettel:
Hah!  When I check the master branch (without the `std::pow()` change) it does indeed break.  

So PR is good.

adamjohansen:
Excellent... any reason not to merge?

tibi77:
Why the `pow` from `src/blockpfgaussianopt.cpp` is not changed? Is there a reason that I am missing? (Asking from a consistency point of view).

eddelbuettel:
@adamjohansen No, I'll do it later.
@tibi77 You can comment on the actual line. Some remain because they are in `arma::pow()`; the confusion (for the Solaris compiler) arose over `std::pow()`

PR #52, opened 2021-05-21
add missing documentation to nonLinPMMH() (or moving documentation to roxygen)

ilyaZar:
Pulled out the discussion from #50, and regarding to @adamjohansen and @LeahPrice , we can discuss to either deal with

1. adding documentation to verbose, msg_freq arguments of nonLinPMMH(), maybe I missed something but I hope it's solved more or less.

2. moving to roxygen2, and if so, I'll get this done within this PR.

In general, moving to roxygen comments should, firstly, be mostly copy and paste from the *.Rd files already present, thus not a big deal, and secondly, we might identify and close missing spots in the documentation.

(obviously, that's not my call at all, just collecting some ideas around here!)

### DIFF
--- a/man/nonLinPMMH.Rd
+++ b/man/nonLinPMMH.Rd
@@ -22,7 +22,7 @@ parameters, the log-prior, the log-likelihood and the MH acceptance rates up to
 the current iteration; defaults to \code{FALSE} in which case only percentage 
 completion of the procedure is printed.}
 
-\item{msg_feq}{Specifies the printing fequency of percentage completion or, if
+\item{msg_freq}{Specifies the printing frequency of percentage completion or, if
 \code{verbose = TRUE}, percentage completion as well as convergence
 diagnostics.}
 
--- a/man/nonLinPMMH.Rd
+++ b/man/nonLinPMMH.Rd
@@ -3,7 +3,7 @@
 \title{Particle marginal Metropolis-Hastings for a non-linear state space model.}
 \usage{
 nonLinPMMH(data, particles = 5000, iterations = 10000, burnin = 0,
-  plot = FALSE)
+  verbose = FALSE, msg_freq = 100, plot = FALSE)
 }
 \arguments{
 \item{data}{A vector of the observed data.}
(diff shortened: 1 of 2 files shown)


eddelbuettel:
FWIW I am aware of at least one existing R package offering to translate Rd to roxygen2 markup for documentation, I have never tried it. I also tend to use roxygen2 in package but have never bothered to translate existing (working !!) Rd files.

eddelbuettel:
The PR is good in the 'obvious fix' sense of correcting the Rd file.  I am a little surprised your computer wants to alter RcppExports again but then I sometimes forget we are on a six-months release cycle with Rcpp and I run pre-release versions :)  I presume you have 1.0.6?

ilyaZar:
Yes `packageVersion("Rcpp")` returns `1.0.6` (and regarding the other remarks on roxygen2, cross referencing and keeping discussions in one place: I read everything carefully, thanks for the information, I do appreciate every bit of it!!).

adamjohansen:
The PR almost does the immediate job; CI is great, and can even spot the typos sometimes:
> Undocumented arguments in documentation object 'nonLinPMMH'
>  ‘msg_freq’
>Documented arguments not in \usage in documentation object 'nonLinPMMH':
>  ‘msg_feq’

If you feel like its worth doing (I don't feel super strongly either way) then file a separate issue for the Roxygen-isation of the documentation and then we won't forget about it, even if it doesn't happen immediately.

adamjohansen:
Looks good to me.

(The last point was obviously a trivial one, but it's well worth scanning through the output of the CI test(s) even when they pass to check for notes or warnings.)

eddelbuettel:
>If you feel like its worth doing (I don't feel super strongly either way) then file a separate issue for the Roxygen-isation of the documentation and then we won't forget about it, even if it doesn't happen immediately.

Seconded.  I also think we can simply have new code use roxygen2 headers in  the R (or C++ file) for documentation. Mix and match is fine, and doing so does not force us to convert the old ones.  So this could be disentangled into two tasks.

eddelbuettel (approved):

eddelbuettel:
(As for the 'line noise' of slightly different RcppExports.cpp created by my newer Rcpp, anyone who wants can run the last release candidate (currently 1.0.6.6)  via   `install.packages("Rcpp", repos="https://RcppCore.github.io/drat")`.  It is very stable and had no commit in a few weeks and very close to what will be 1.0.7 in July.  Don't feel forced though, just mentioning it as it is in fact easy to do ...)

PR #56, opened 2021-06-22
Add facilities that allow for tracking of ancestral lines

ilyaZar:
So here are the discussed changes to allow for tracking of ancestral lines (just squashed my commits for the first time so the commit history will be cleaner than last time :) ). 

I agree with @adamjohansen about the documentation; we have the comments to the include files and the .Rd files for functions that are meant to be called via R or am I mistaken?

@eddelbuettel I upgraded `Rcpp` to the Drat version, so this time no changes to the export-files I think.

Hope the CI passes this time, I'll check this first thing in the morning!

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2021-06-23  Ilya Zarubin  <ilya.a.zarub@gmail.com>
+
+	* inst/include/history.h: add facilities for tracking of ancestral lines
+	(mostly additions to the historyelement class)
+	* inst/include/sampler.h: add computation of ancestral lines making us of
+	updated facilities from the historyelement class; conditional switch
+	allowing for an additional HistoryType::AL
+
 2021-05-21  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (URL, BugReports): Added
@@ -47,7 +55,7 @@
 
 2021-01-12  Dirk Eddelbuettel  <edd@debian.org>
 
-        * .github/workflows/ci.yaml: Add CI runner using r-ci
+    * .github/workflows/ci.yaml: Add CI runner using r-ci
 	* README.md: Add new badge
 
 2020-08-29  Dirk Eddelbuettel  <edd@debian.org>
--- a/inst/include/sampler.h
+++ b/inst/include/sampler.h
@@ -62,7 +62,8 @@ namespace PathSamplingType
 namespace HistoryType
 {
     enum Enum { NONE = 0,
-        RAM};
+        RAM,
+        AL};
 }
 
 namespace smc {
@@ -161,6 +162,16 @@ namespace smc {
         historyflags GetHistoryFlags(long n) const {return History[n].GetFlags();}
         /// Returns the Effective Sample Size of the specified particle generation.
         double GetESS(long n) const {return History[n].GetESS();}
+        /// Returns a vector of population objects re-ordered to match ancestral lines of the particle system
+        std::vector<population<Space>> GetAPop(void) const;
+        /// Returns indices for ancestral line of particle 'n'
+        arma::Col<unsigned int> GetALineInd(long n) const;
+        /// Returns ancestral line of particle 'n'
+        std::vector<Space> GetALineSpace(long n) const;
+        /// Returns current iteration resampled indices
+        arma::Col<unsigned int> GetuRSIndices(void) const {return uRSIndices;}
+        /// Returns current iteration resampled index number 'n'
+        int GetuRSIndex(long n) const {return uRSIndices(n);}
         ///Returns the history number of MCMC iterations performed during this iteration.
         int GetHistorymcmcRepeats(long n) {return History[n].mcmcRepeats();}
         ///Returns the additional algorithm parameters.
@@ -403,7 +414,67 @@ namespace smc {
     {
         return expl(2*stableLogSumWeights(pPopulation.GetLogWeight())-stableLogSumWeights(2.0*pPopulation.GetLogWeight()));
     }
+    template <class Space, class Params>
+    std::vector<population<Space>> sampler<Space,Params>::GetAPop(void) const
+    {
+        long N = GetNumber();
+        long historyLength = GetHistoryLength();
+        /// Output container for particle population ancestor lines
+        std::vector<population<Space> > populationAL(historyLength);
+        /// Get final iteration indices
+        arma::Col<unsigned int> tmpAI = History[historyLength-1].GetAIndices();
+        /// Get final iteration particle values
+        populationAL[historyLength-1] = GetHistoryPopulation(historyLength-1);
+        /// Iterating from last iteration to initiliaization period whereby:
+        /// 1. Re-ordering particle population elements for each t=0,...,T-1 accoridng to ancestral lines
+        /// 2. Storing these particle population classes in output: populationAL
+        if (historyLength > 1){
+            for (int t = historyLength - 2; t >= 0; --t){
+                for(int i = 0; i < N; ++i){
+                    /// Setting each particle population element at indices i=0,...,N-1 to its ancesor index in the output vector populationAL
+                    populationAL[t].SetValueN(GetHistoryPopulation(t).GetValueN(static_cast<int>(tmpAI(i))), i);
+                }
+                /// Re-mapping ancestor indices to retrieve ancestor indices a period before
+                tmpAI = (History[t].GetAIndices()).elem(tmpAI);
+            }
+        }
+        /// The output container stores population classes re-ordered according to their ancestor lines for t=0,...,T-1. So, for example, the first each element of populationAL[t] is the direct ancestor population class element of populationAL[t+1].
+        return populationAL;
+    }
+    template <class Space, class Params>
+    arma::Col<unsigned int> sampler<Space, Params>::GetALineInd(long n) const
+    {
+        long N = GetNumber();
+        long historyLength = GetHistoryLength();
 
+        /// Get final iteration indices and final iteration particle values
+        unsigned int tmpAI = (History[historyLength-1].GetAIndices()).at(static_cast<int>(n));
+        arma::Col<unsigned int> ALineInd(historyLength);
+        ALineInd[historyLength - 1] = static_cast<int>(n);
+
+        for(int t = historyLength - 2; t >= 0; --t){
+            tmpAI = (History[t].GetAIndices()).at(tmpAI);
+            ALineInd[t] = tmpAI;
+        }
+        return ALineInd;
+    }
+    template <class Space, class Params>
+    std::vector<Space> sampler<Space,Params>::GetALineSpace(long n) const
+    {
+        long N = GetNumber();
+        long historyLength = GetHistoryLength();
+
+        std::vector<Space> ALineSpace(historyLength);
+
+        /// Get final iteration indices and final iteration particle values
+        arma::Col<unsigned int> ALineInd(historyLength);
+        ALineInd = this->GetALineInd(static_cast<int>(n));
+
+        for(int t = 0; t < historyLength; ++t){
+            ALineSpace[t] = GetHistoryPopulation(t).GetValueN(static_cast<int>(ALineInd(t)));
+        }
+        return ALineSpace;
+    }
     /// At present this function resets the system evolution time to 0 and calls the moveset initialisor to assign each
     /// particle in the ensemble.
     ///
@@ -441,6 +512,9 @@ namespace smc {
         }
         else {
             nResampled = 0;
+            if(HistoryType::AL) {
+                uRSIndices = arma::linspace<arma::Col<unsigned int>>(0, N - 1, N);
+            }
             pAdapt->updateForMCMC(algParams,pPopulation,acceptProb,nResampled,nRepeats);
         }
 
@@ -457,13 +531,22 @@ namespace smc {
         pAdapt->updateEnd(algParams,pPopulation);
 
         //Finally, the current particle set should be appended to the historical process.
-        if(htHistoryMode != HistoryType::NONE) {
+        if(htHistoryMode != HistoryType::NONE){
             History.clear();
             historyelement<Space> histel;
-            histel.Set(N, pPopulation, nAccepted, nRepeats, historyflags(nResampled));
+            switch(htHistoryMode) {
+            case HistoryType::RAM:
+                histel.Set(N, pPopulation, nAccepted, nRepeats, historyflags(nResampled));
+                break;
+            case HistoryType::AL:
+                histel.Set(N, pPopulation, nAccepted, nRepeats, historyflags(nResampled), uRSIndices);
+                break;
+            /// To avoid compiler warnings, HistoryType::NONE is handled
+            case HistoryType::NONE:
+                break;
+            }
             History.push_back(histel);
         }
-
         return;
     }
 
@@ -641,6 +724,9 @@ namespace smc {
         }
         else {
             nResampled = 0;
+            if(HistoryType::AL) {
+                uRSIndices = arma::linspace<arma::Col<unsigned int>>(0, N - 1, N);
+            }
             pAdapt->updateForMCMC(algParams,pPopulation,acceptProb,nResampled,nRepeats);
         }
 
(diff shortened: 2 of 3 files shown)


ChangeLog
```diff
@@ -1,3 +1,8 @@
+2021-05-20  Ilya Zarubin  <ilya.a.zarub@gmail.com>
```
LeahPrice, line 1: Perhaps we should update the date to avoid confusion when looking back at discussions.

ChangeLog
```diff
@@ -1,3 +1,8 @@
...
+
+	* inst/include/history.h: add facilities for tracking of ancestral lines (mostly additions to the historyelement class)
```
LeahPrice, line 3: We've been restricting the line lengths to I think around 80 characters/columns for readability, so these two dot points could be split across more lines. E.g. see your 2021-05-20 src/nonLinPMMH.cpp item.

Adam and Dirk use Emacs where this is all automatic.

LeahPrice (commented):
These changes look good to me, as discussed in #55, but I have a couple of formatting suggestions for the changelog.

ChangeLog
ilyaZar: Done, but I squashed my branch commits and `git push origin +historyelement-class-additions-for-AL ` hence the outdated label (also below)...

ilyaZar:
@LeahPrice thanks for the remarks, all done!

inst/include/sampler.h
```diff
@@ -161,6 +162,15 @@ namespace smc {
         historyflags GetHistoryFlags(long n) const {return History[n].GetFlags();}
         /// Returns the Effective Sample Size of the specified particle generation.
         double GetESS(long n) const {return History[n].GetESS();}
+        /// Returns a vector of population objects re-ordered to match ancestral lines of the particle system
+        std::vector<population<Space>> GetApop(void) const;
```
adamjohansen, line 166: This is rather trivial, but should this be `GetAPop` or similar for consistency with the other use of A in method names around here?

inst/include/sampler.h
```diff
@@ -161,6 +162,15 @@ namespace smc {
         historyflags GetHistoryFlags(long n) const {return History[n].GetFlags();}
         /// Returns the Effective Sample Size of the specified particle generation.
         double GetESS(long n) const {return History[n].GetESS();}
+        /// Returns a vector of population objects re-ordered to match ancestral lines of the particle system
+        std::vector<population<Space>> GetApop(void) const;
+        // void GetALineInd(long n) const;
+        arma::Col<unsigned int> GetALineInd(long n) const;
+        std::vector<Space> GetALineSpace(long n) const;
+        /// HELPER FUNCTION THAT MAY BE DELETED; RETURNS uRSIndices
```
adamjohansen, line 170: Are these functions intended to be used externally? If not should they be private or protected?

It just seems a bit unfriendly to out to the interface something with a "MAY BE DELETED" comment attached. I don't immediately see any harm in leaving these functions but if we do I'd be inclined to do that with the view to keeping them permanently (or as permanently as one does anything in code).

inst/include/sampler.h
```diff
@@ -403,7 +413,71 @@ namespace smc {
...
+        /// Get final iteration indices
+        arma::Col<unsigned int> AI_tmp = History[historylength-1].GetAindices();
+        /// Get final iteration particle values
+        populationAL[historylength-1] = GetHistoryPopulation(historylength-1);
+        /// Iterating from last iteration to initiliaization period whereby:
+        /// 1. Re-ordering particle population elements for each t=0,...,T-1 accoridng to ancestral lines
+        /// 2. Storing these particle population classes in output: populationAL
+        if (historylength > 1){
+            for (int t = historylength - 2; t >= 0; --t){
+                for(int i = 0; i < static_cast<int>(N); ++i){
```
adamjohansen, line 433: This looks like a bad idea.

If we use a long for `N` to allow for large particle sets on architectures with small words or similar then we should really be using a long for indices which iterate over the particle set shouldn't we?

If there's a reason to strong prefer int then there's probably no harm in changing the type of N from long to int given the architectures and compilers in current used. There aren't many 16 bit `int`s around any more.

inst/include/sampler.h
```diff
@@ -403,7 +413,71 @@ namespace smc {
     {
         return expl(2*stableLogSumWeights(pPopulation.GetLogWeight())-stableLogSumWeights(2.0*pPopulation.GetLogWeight()));
     }
+    template <class Space, class Params>
+    std::vector<population<Space>> sampler<Space,Params>::GetApop(void) const
+    {
+        long N = GetNumber();
+        long lHistorylength = GetHistoryLength();
+        int historylength = static_cast<int>(lHistorylength);
```
adamjohansen, line 421: Is there a reason to keep casting long to int? In the abstract losing bits like this seems a bad idea, although `long` and `int` may not be different on some architectures.

inst/include/sampler.h
```diff
@@ -403,7 +413,71 @@ namespace smc {
...
+    {
+        long N = GetNumber();
+        long lHistorylength = GetHistoryLength();
+        int historylength = static_cast<int>(lHistorylength);
+
+        std::vector<Space> AlineSpace(historylength);
+
+        /// Get final iteration indices and final iteration particle values
+        arma::Col<unsigned int> AlineInd(historylength);
+        AlineInd = this->GetALineInd(static_cast<int>(n));
```
adamjohansen, line 474: Again there seems to be some inconsistency around case with these As. I don't want to obsess over this, but it does jar a bit when the function is `GetALineInd` and the variable is `AlineInd`.

inst/include/sampler.h
```diff
@@ -441,6 +515,8 @@ namespace smc {
         }
         else {
             nResampled = 0;
+            uRSIndices = arma::zeros<arma::Col<unsigned int> >(static_cast<int>(N));
```
adamjohansen, line 518: I don't use armadillo enough, I suspect, but is there a reason to first create and initialise to zero and _then_ to fill with specified values? It naively seems a touch redundant.

inst/include/sampler.h
```diff
@@ -457,13 +533,21 @@ namespace smc {
...
             History.clear();
             historyelement<Space> histel;
-            histel.Set(N, pPopulation, nAccepted, nRepeats, historyflags(nResampled));
+            switch(htHistoryMode) {
+            case HistoryType::RAM:
+                histel.Set(N, pPopulation, nAccepted, nRepeats, historyflags(nResampled));
+                break;
+            case HistoryType::AL:
+                histel.Set(N, pPopulation, nAccepted, nRepeats, historyflags(nResampled), uRSIndices);
+            /// To avoid compiler warnings, HistoryType::NONE is handled
```
adamjohansen, line 545: Don't you want a `break` here? Rather than falling through to the NONE case?

inst/include/sampler.h
```diff
@@ -641,6 +725,8 @@ namespace smc {
         }
         else {
             nResampled = 0;
+            uRSIndices = arma::zeros<arma::Col<unsigned int> >(static_cast<int>(N));
```
adamjohansen, line 728: I guess we could do this conditionally on wanting to store this information, but I can see the case for keeping things simple.

adamjohansen (changes requested):
This looks good. There are a few details that I've commented on.

Aside from cases of bits in names (sorry, this is a pain but it seems like a good time to try to do something consistent here) the main thing that is bothering me is that we seem to do a lot of casting between `long` and `int` which although it will probably _just work_ for most people on most modern systems seems to me to be a bad thing in the abstract. This might reflect deficiencies in the existing code base that this PR was just trying to work around, but is it feasible to avoid needing to do it without undue effort or are there just too many moving parts?

(If it's necessary to have `int` rather than `long` in places for any reason then it would be arguably more sensible to replace the `long`s with `int`s than to mix the two.)

adamjohansen:
@ilyaZar we do have the documentation that you mention, but I've had it in mind that comments on the code together with man-page level documentation of some R functions which are essentially example implementations isn't really enough for someone to come to the code cold and implement their favourite algorithm with it.

Consequently, we ought when we have time and capacity to provide some more substantial documentation. We were, at one stage, trying to put together an analogue of [this SMCTC overview](http://www.jstatsoft.org/v30/i06/paper), with the aim of telling people roughly how the library works and how to us it along with a few examples but we never quite found the time to finish it.

At some point I think something like that is probably going to be necessary to make things accessible to the people who might find the code useful. It could just be a more detailed user manual than we currently have, or something like the SMCTC summary, but a fairly holistic description of how to go from nothing to a working SMC algorithm seems to be needed.

inst/include/sampler.h
eddelbuettel: Yes, looking at 518 and 519 I think only 519 ("linspace") is needed.

Also, Arma has `typedef` for the common type so instead of arma::icovec and arma::ivec are the same as a arma::Col<unsiged int>.  I guess it takes use to make it more legible as it is new -- arma::Col<unsigned int> is understood by everybody but a mouthful.

inst/include/sampler.h
eddelbuettel: I also find `uint64_t and `int64_t` more obvious, I think C++ pros also like size_t for the indices. 

```r
> Rcpp::evalCpp("sizeof(long)")    
[1] 8 
> Rcpp::evalCpp("sizeof(int)") 
[1] 4   
> Rcpp::evalCpp("sizeof(int64_t)")      
[1] 8  
> Rcpp::evalCpp("sizeof(size_t)")    
[1] 8
> 
```

inst/include/sampler.h
ilyaZar: 1. I think that depends on how we want to implement conditional SMC and grant the user access to these class members. 
* It would be more flexible to allow the user access to all trajectories, hence I'd suggest `GetALineSpace()` should be public.
* I believe that `GetALineSpace()` is only ever useful for internal purposes, specifically, also for derived cond. SMC. So `GetALineInd()` should be protected. 

If no objections I'll do as above.

2. Ah I see, my mistake, just wanted to highlight this with capslock for a discussion.
* I think, we might want to decide whether to make ancestral lines accessible only via the historyelement class (via `historyelement.SetAindices()`) or additionally (but this might be an unnecessary deflection since it only gives access to the current iteration resampled indices) via 
https://github.com/rcppsmc/rcppsmc/blob/6bd4ceaa976d9285cb1d867515bb003e0f3d7361/inst/include/sampler.h#L170-L171

* The same goes for 
https://github.com/rcppsmc/rcppsmc/blob/6bd4ceaa976d9285cb1d867515bb003e0f3d7361/inst/include/sampler.h#L172-L173

Maybe there are other merits to these functions, that I don't see. In general, I'd be a bit more inclined to delete unnecessary facilities (only keeping what's relevant, for clarity purposes). 

What do the others think @eddelbuettel @LeahPrice ?

inst/include/sampler.h
ilyaZar: Done, will keep in mind for next time!

inst/include/sampler.h
ilyaZar: Done, will keep in mind for next time!

inst/include/sampler.h
ilyaZar: Done!

inst/include/sampler.h
ilyaZar: At first, an `if()` looked a bit ugly to me, so I deleted it for the moment. 

Now I believe, an external person might get confused by this redundancy in case of `HistoryType::RAM` so I added it now.

inst/include/sampler.h
ilyaZar: 1. Got rid of the unnecassry line.

2. On typedefs: agreed @eddelbuettel  , but I followed the existing code base like here:
* https://github.com/rcppsmc/rcppsmc/blob/6bd4ceaa976d9285cb1d867515bb003e0f3d7361/inst/include/sampler.h#L251-L255
* https://github.com/rcppsmc/rcppsmc/blob/6bd4ceaa976d9285cb1d867515bb003e0f3d7361/inst/include/sampler.h#L796

Maybe we should invest a bit of time and do a PR for re-naming everything to pure typedefs?

ilyaZar:
Allright. So before I do the my re-work of the PR:

1. Using typdef or explicit naming (above, I adhear to the existing explicit naming `arma::Col<unsigned int>` in the current codebase, but if we wish to be consistent then we probably ought to change it everywhere and not just for this particular PR?)
2.  @adamjohansen : the `long` vs. `int` is the same issue, I think it might be wise to decide now and do a PR fixing this (but not just for this commit).

I believe it's worth a "discussion" (or at least 1-2 seperate PR unrelated to this one because this one deals with the ancestral lines solely) because there are numerous incidents of 1./2. inconsistencies that we might want to address.

inst/include/sampler.h
ilyaZar: Changed!

inst/include/sampler.h
eddelbuettel: Maybe.  We should pick a consistent style. I can see reasons for being explicit (everybody understands) and for being shorter (but maybe that is only used for more Armadillo-heavy projects?)

inst/include/sampler.h
adamjohansen: I'll interrupt given I raised this in the first place.

Presumably in 1. the `GetALineSpace` in the second bullet point is a typo for `GetALineInd`. No objections from me, although I didn't want to insist on anything just that if we have functions which we might get rid of them exposing their interface might not be helpful. I still have no problem with keeping them; it's difficult to predict every possible use case and making everything available to users that might want it doesn't seem to hurt.

No strong views either way on 2. Largely a matter of taste, I think.

inst/include/sampler.h
adamjohansen: No strong views either way; but consistent is good.

inst/include/sampler.h
adamjohansen: I don't have strong views, it just seems like an unnecessary operation if the information isn't wanted and I still remember when computers were slow...

adamjohansen:
Thanks, @ilyaZar 
1. No strong views from me; the person writing the code tends to get to make these decisions unless someone has a reason to favour one or the other (@edelbuettel is definitely the armadillo expert...).
2. I was being lazy and hadn't checked whether we were doing this already and you'd been pushed into following what we were doing. If that's the case then it would probably make sense to resolve this PR without worrying about that and add an issue ticket to fix this throughout the code. If the existing code avoids this then we might as well keep things clean here.  A quick `grep static_cast *.h` does seem to reveal existing instances of this. It's not clear to me if this is because of an interaction with armadillo or if we've just been sloppy in the past, but it does seem that it's a wider issue than this PR and should probably be dealt with separately.
3. Examples, vignettes etc. are all good. My feeling is that a longer form more manual-style piece of documentation would still be of value (and we have something half written from memory), but that's not a good reason not to supplement it with other things.

inst/include/sampler.h
LeahPrice: I also don't think we should make functions public unless we plan on keeping them, and I'm not too concerned about keeping functions that we can't see a direct use for right now if they might be useful later. I don't have any strong opinions on point 2.

ilyaZar:
Ok so the last changes are what has been asked for above, plus

1. I left the long form i.e. `arma::Col<unsigned int>`, not using a typedefs; but I added typedefs into the new discussion to do list #57 
2. I changed to `long` leaving a major revision of `int` vs `long` for later; also added this to the discussion To-Do list #57 
3. `history.h`: changed names form `SetAindices()` to `SetAIndices()` and from `GetAindices()` to `GetAIndices()` as I feel this also is more in line with our style of naming (but has not been asked for explicitly)

adamjohansen:
That all looks good to me.

LeahPrice (approved):

adamjohansen (approved):
Sorry for holding things up, my comment in the main discussion that it all looked good was made after looking at the code I should probably have been a bit more formal about it.

adamjohansen:
The forced push seems to have confused the history of this discussion a bit but everything looks good to me and, by the look of it, @LeahPrice. Any objections, @eddelbuettel?

eddelbuettel (approved):
LGTM

PR #60, opened 2021-08-15
Conditional SMC

ilyaZar:
The conditional SMC branch from the discussion.


src/RcppExports.cpp
```diff
@@ -7,6 +7,11 @@
 
 using namespace Rcpp;
 
+#ifdef RCPP_USE_GLOBAL_ROSTREAM
+Rcpp::Rostream<true>&  Rcpp::Rcout = Rcpp::Rcpp_cout_get();
+Rcpp::Rostream<false>& Rcpp::Rcerr = Rcpp::Rcpp_cerr_get();
+#endif
+
```
ilyaZar, line 14: Hi @eddelbuettel. Sorry for not checking RcppExports.cpp for that particular commit (remember that this was a thing last time). 

It think it appeared because of my use of `Rcpp::Rcout` to do some debugging thing and then I forgot to remove print statements before committing (as I usually do) to have a clean commit.

I think it probably appears via compile attributes
- `R CMD INSTALL --preclean --no-multiarch --with-keep.source` is executed directly in vscode whenever I build rcppsmc there ->lines 10-13 do NOT appear
- "Clean and rebuild in RStudio" from reading the RStudio pane: 
   ```
   ==> Rcpp::compileAttributes()

   * Updated src/RcppExports.cpp
   * Updated R/RcppExports.R

   ==> R CMD INSTALL --no-multiarch --with-keep.source rcppsmc
   ```

So the latter specifically adds those lines while my vscode build execution does not (compile attributes is missing).

Now here is the thing (why this might be worth mentioning):
1. Now if comment out that particular statement like so `// Rcpp::Rcout <<....` and rebuild neither RStudio nor vscode removes that entry in RcppExports.cpp.
2. If I delete that (commented out) line alltogether neither RStudio nor vscode remove the entry in RcppExports.cpp

adamjohansen:
Hi @ilyaZar, I'm starting to lose track of where we are.

Is there stuff you still want to do before we consider this "finished" or should I have a final look at the overall PR now?

ilyaZar:
Hi @adamjohansen ! 

Sorry, you are absolutely right, I probably should have said what my plans are before doing a random walk through the library and  changing and committing things that may not be necessary.

I want to do a small change int the main body of the `conditionalSampler` and check on the residual resampling part of the `conditionalResampling` function a final time, going through it step by step. I'll let you know when I am done (probably tommorrow afternoon)!

Thanks!

src/RcppExports.cpp
eddelbuettel: (That is a non issue and simply due to Rcpp 1.0.7 being newer than 1.0.6 or other. It's a minor new feature for the output device Rcpp uses -- see the release notes for Rcpp 1.0.7.  In short nothing to worrry about.   [ Unless I completely misunderstand what you are asking. ] )

eddelbuettel:
I am not a fan of PR that cover a dozen files in whitespace-only changes.

ilyaZar:
Sorry, I was in the middle of writing some remarks in the discussion and I missed this message in the process @eddelbuettel . I'll have a thought how to do revert this (removing whitespaces helped me a bit with my editor behaviour but I realize now that it's on me to make my vscode not highlighting all the extra whitespaces and that this does not really belong into a committ...)

eddelbuettel:
Yeah if you could tame vscode for now (as Adam and I do with The Editor That Matters (TM) ;-) ) and maybe resubmit it would be cleaner.  It is clearly on us -- should not have those back then when -- but there is also a general sentiment that _bulk whitespace commits_ are meh.

ilyaZar:
Hi, Done: just reverted to a prior commit and forced pushed only relevant changes.

I hope I did not mess around with your git-histories @adamjohansen @LeahPrice  @eddelbuettel. From what I could find out, one should __*not*__ `git rebase` and force push because rebasing changes previously written commit objects in git histoy(I __think__).

I believe that caused the problem @adamjohansen had last time when merging my pull request because I did exactly that...

It may be different if one `git revert`, then apply only relevant changes and then force push (as @eddelbuettel suggested) but I could not really find out more about this ... So sorry if this causes any problems in your local branches or annoying merge conflicts...

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
+// -*- mode: C++; c-indent-level: 4; c-basic-offset: 4; indent-tabs-mode: nil; -*-
```
eddelbuettel, line 1: This could be removed.  It's how Adam and I (and many others....) used to tell Emacs about preferences on per-file basis.

These days we have the file `.editorconf` -- which does this project-wide, and for other editors.

eddelbuettel (approved):
Looks good to me (on a somewhat superficial skim).

Thanks for removing the whitespace-only changes, it's appreciated.  We should be fine pulling once merged.

inst/include/conditionalSampler.h
ilyaZar: Agreed! I think I'll do it in #61 and then probably for all files (it's something I noticed in all files) ?

@adamjohansen do you need these type of lines for something?

adamjohansen: No, it's an artifact of the past as @eddelbuettel suggested. Sorry for being slow, I will look at things properly later today.

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+//
+// sampler.h: Rcpp integration of SMC library -- sampler object
```
adamjohansen, line 3: A pedantic point, but shouldn't this be conditionalSampler.h with a slughtly more descriptive name?

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+// RcppSMC is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with RcppSMC. If not, see <http://www.gnu.org/licenses/>.
+
+//! \file
+//! \brief Defines the overall sampler object.
```
adamjohansen, line 23: It would be good for the one-line summary of what the file does to be different to that for sampler.h

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+            ///Perform conditional iterations until the specified evolution time is reached
+            void IterateUntil(long lTerminate);
+            ///Resample the particle set using the specified resampling scheme specifically adjusted for conditional resampling
+            void conditionalResample(ResampleType::Enum lMode);
+            ///Move the reference particle by setting the coordinate to the reference value and re-weighting.
+            void MoveReferenceParticle(void);
+            ///Dump a specified particle to the specified output stream in a human readable form
+            std::ostream & StreamParticle(std::ostream & os, long n) const;
+            ///Dump the entire particle set to the specified output stream in a human readable form
+            std::ostream & StreamParticles(std::ostream & os, int digits) const;
```
adamjohansen, line 142: I'm not proposing that we change anything in this PR, but we should have consistent interfaces for `sampler` and `conditionalSampler` insofar as we can, so we should add an issue/todo item somewhere to make the printing interfaces consistent. Whether the number of digits to print belongs within an individual sampler object or more globally isn't immediately obvious to me and I don't have strong views, but if it's handled by the conditional case it should also be handled by the base case, I would think.

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+            friend std::ostream& operator << <>(std::ostream &, const conditionalSampler<Space,Params>&);
+
+            ///Throws exception when adaptation related members of the base sampler class are used:
+            void SetAdaptMethods(adaptMethods<Space,Params>* adaptMethod) {throw SMC_EXCEPTION(CSMCX_USING_ADAPTATION, "Adaptation methods not supported for conditional sampler class.");}
```
adamjohansen, line 146: I'd personally be inclined to add the word currently before supported in these messages; in a sense most of these are things we haven't implemented because they're complicated and time is precious rather than because we've made a conscious decision not to.

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+
+        switch(lMode) {
+        case ResampleType::MULTINOMIAL:
+        default:
+            {
+                //Algorithm 3
+                //Step 0:
+                //Sample conditional index K_t from appropriate version of the "lambda" distribution i.e. uniformly on {1,...,N} in case of Multinmial resampling.
+                // long Kt = floor(unif_rand()*static_cast<double>(N)); // sample lamba(k_{t}|w_{t-1}, k_{t-1})=1/N
+                long Kt = 0;
```
adamjohansen, line 388: What is done here doesn't seem to match the comment.

inst/include/smc-exception.h
```diff
@@ -38,7 +38,14 @@
 #define SMCX_MISSING_HISTORY 0x0010
 ///Exception thrown if an attempt is made to instantiate a class of which a single instance is permitted more than once.
 #define SMCX_MULTIPLE_INSTANTIATION 0x1000
-
+///Exception thrown if an attempt is made to instantiate a class of which a single instance is permitted more than once.
+#define SMCX_MULTIPLE_INSTANTIATION 0x1000
+//Exception thrown if an attempt is made to instantiate a class of which a single instance is permitted more than once.
+#define SMCX_MULTIPLE_INSTANTIATION 0x1000
```
adamjohansen, line 44: There seems to be some duplication here? (41-42; 43-44 look rather similar up to a /).

inst/include/smc-exception.h
```diff
@@ -38,7 +38,14 @@
...
+//Exception thrown if an attempt is made to use facilities of the base sampler class related to MCMC moves.
+#define CSMCX_USING_MCMC 0x1000
```
adamjohansen, line 46: It would be good to define unique codes for each type of exception so that catching routines can differentiate between difference problems.

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+                    uRSIndices.at(Kt) = referenceTrajectoryIndices.at(T);
+                }
+                //else already assigned under 1.
+                break;
+            }
+        }
+        //Perform the replication of the chosen.
+        for(int i = 0; i < N ; ++i) {
+            if(uRSIndices(i) != static_cast<unsigned int>(i)){
+                pPopulation.SetValueN(pPopulation.GetValueN(static_cast<int>(uRSIndices(i))), i);
```
adamjohansen, line 604: I may be misunderstanding something here, but is there a possible race condition here? The value of each particle is set to the value of its ancestor in a loop but the value of the ancestor could already have been overwritten, no? (This was dealt with in the standard case by assuring that if particle i had any offspring then one of them had index i but I don't think that this is the case here.)

Am I missing something (I'm quite prepared to believe that I am)? Otherwise the cleanest option is to copy the pre-resampling particle set and use the values within that collection where we have `GetValueN` at present.

adamjohansen (commented):
Thanks for this @ilyaZar -- I've just read through all of this and still have a few small comments, one in particular about exactly how you're managing the act of setting particle values within the conditional resamping that didn't seem completely clear to me. But this is very close to being done, I think.

inst/include/conditionalSampler.h
ilyaZar: Done!

inst/include/conditionalSampler.h
ilyaZar: Done! I think I understood the point about the first line so I won't remove it (for consistency with the other files in the package).

inst/include/conditionalSampler.h
ilyaZar: Yes indeed! Added this to the to-do list of GSoC. We might want to make an overall To-Do in the discussion section not related to GSoC for future work that also includes past To-Dos (including or reference #31 bullet point 4 here makes sense).

inst/include/conditionalSampler.h
ilyaZar: Done! But I changed the description of the conditionalSampler class line l. 25 of `conditionalSampler.h` to provide a bit more information and documentation for the user. 

It's probably best if you have a look at the exact wording of this line to make that statement as precise and clear as it needs to be (it's a language thing and my english is clearly only my second best :) )

inst/include/conditionalSampler.h
ilyaZar: Added comment.

 I changed this part in the last commit to better see if the conditional conditional trajectory is set correctly and also to save one draw. I'll leave 386 and 387 there for the moment and add an additional line above 388?

inst/include/smc-exception.h
ilyaZar: Done!

inst/include/smc-exception.h
ilyaZar: Done.

inst/include/conditionalSampler.h
ilyaZar: Yes you are absolutely right! 

I've done a change though I am not sure if it is 100% what you suggested. Copying the whole particle might be too much (and also It think as currently implemented `=` does a shallow copy of the pPopulation object).

So I copied only `pPopulation.value` member via `GetValue()` into a seperate `populationValueCopy` and did the ancestor assignment with values taken from that vector.

We might want to add a note on the To-Do:
1. Either add a member that permutes the particle set to the `population` class (and then implement it efficiently later). The operation might be common enough to justify a member-function for that.
2. Permutation member function ccould be optimized. I __*think*__ there are some algorithmic ways how to permute without pre-copying that might be worth considering for large particle sets, but that is definitely something for the future too.

ilyaZar:
Hi all,

quick sorry @eddelbuettel . I wanted to ask @adamjohansen  for reviewing but ended up hitting the wrong button, hence the unnecessary re-request.

Seems like I can't undo it once the request is out...

Are there still changes you'd like me to add or test something? (I am essentially free this weekend)

If not, I'll tend to finalize and write up everything (in terms of commits made and other things to put on the website) for GSoC now (better now then on the deadline next monday...) and we can discuss the remaining issues in the next and following weeks?

inst/include/conditionalSampler.h
adamjohansen: Thanks. Indeed, I meant copy the collection of values rather than the object itself. 
I agree that 1, 2 might make sense -- and that we at least want the facility to perform deep copying, whether that's default behaviour (which perhaps it should be) or not.

inst/include/conditionalSampler.h
adamjohansen: Thanks; the main point was just to avoid a discrepancy between what was described and what was actually done... even to us that would be confusing in the end.

inst/include/conditionalSampler.h
adamjohansen: Thanks. It might well be a good idea to have a more structured lto do list somewhere.

inst/include/conditionalSampler.h
adamjohansen: I think what you have here should be fine; but perhaps the copy should be explicitly cleared once we're finished with it?

adamjohansen (approved):
I think we're essentially there; one query about whether the vector of copied particle values ought to be explicitly cleared after use is all that remains and that's not critical.

adamjohansen:
Is there an example that uses this code available to the world at large? That's the one thing that it would obviously be good to provide; e.g. a simple linear gaussian setting in which one could compare the various implementations (perhaps via normalizing constant estimates which are pretty well understood).

adamjohansen:
@LeahPrice does this look good to you?

ChangeLog
```diff
@@ -1,3 +1,79 @@
...
+	calls to not use parameters related to adaptation or MCMC-moves
+	* inst/include/history.h: overload histel.Set() s.th. setting of data
+	members that are related to adaptation or MCMC-move facilities is prohibited
+	* inst/include/smc-exception.h: add exception macros for exceptions related
+	to calls of redundant methods (see first bullet point)
+
+2021-08-11  Ilya Zarubin  <ilya.a.zarub@gmail.com>
+
+	* inst/include/conditionalSampler.h: resolve nondependent name look-up in
+	derived conditionalSampler class via using-statements instead of explicit naming to make code cleaner and facilitate readability (implicit name
```
eddelbuettel, line 23: Line length could do with adjustment.

eddelbuettel (approved):
Looks fine, one trivial comment.

LeahPrice:
Thanks everyone. I read through all of the code and made a couple of very small comments last night. I can't say I understand algorithm 6 completely, but I'm happy with the PR and ready for it to be merged when others are

LeahPrice (approved):

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+                Rcpp::NumericVector lambdaWeightsStratified = Rcpp::wrap(strataWeights/dRSWeightsCumulative.at(referenceTrajectoryIndices.at(T)));
+                //    0.5. Sample K_t from 0.4.
+                long Kt = Rcpp::sample(N, 1, false, lambdaWeightsStratified)[0] - 1;
+                //Step 1:
+                //Connect the "chosen" ancestor index to previous reference trajectory: A_{t - 1}^{K_t} = K_{t - 1}
+                uRSIndices.at(Kt) = referenceTrajectoryIndices.at(T);
+                //Step 2:
+                //Calculation of empirical distribution function F_{t - 1}^N(i) is done and equal to computation of cumulative normalized weights stored in dRSWeightsCumulative.
+                //Step 3: Generate ancestor indices and sssign to offspring indices {1,...,N}\{Kt}.
+                std::vector<unsigned int> tmpIterator(N - 1);
```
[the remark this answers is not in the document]
LeahPrice: Sorry if this and the other comment didn't come through last night (20th) - I didn't realise I had to select approve before these would be visible

ilyaZar:
* > Is there an example that uses this code available to the world at large? That's the one thing that it would obviously be good to provide; e.g. a simple linear gaussian setting in which one could compare the various implementations (perhaps via normalizing constant estimates which are pretty well understood).
 
    @adamjohansen yes! The SVmodelRcppSMC package is implementing PG. I added a link to the package in PR #61 in the  README file (where the other packages are listed). I guess we can provide an "example use case to the world" whenever we decide that this PR is finished. (simple linear Gaussian setting as an explicit package example is certainly also doable if you fancied that let me know!)

* I feel like this is a really stupid question on my side but why do you want to clear `populationValueCopy` (the vector of particle values) explicitly? It's a local variable to the function (call) `conditionalResample()` and should be destroyed (is destroyed or deleted from the stack what you mean by "explicitly cleared"?) automatically once the function returns? I found https://stackoverflow.com/questions/35514909/how-to-clear-vector-in-c-from-memory and based on the post from Johan Lundbeg `populationValueCopy.clear()` would do the job, yet it gets "cleared/destroyed" right after the loop, or?

* @LeahPrice  thanks for pointing out these cases! (I could swear we already had the `N-1` vs `N` thing, but strangely, though, that this keeps hunting me :) )

inst/include/conditionalSampler.h
```diff
@@ -0,0 +1,673 @@
...
+                dRSWeights = exp(pPopulation.GetLogWeight() - stableLogSumWeights(pPopulation.GetLogWeight()));
+                //Step 1:
+                //Assign deterministic offpring indices for each i={1,...,N}.
+                for (int i : tmpIterator) {
+                    //Compute (integer part of) expected number of offspring
+                    expectedNumberOffspring = std::floor(N * dRSWeights.at(i));
+                    //Generate D_i set:
+                    if(expectedNumberOffspring > 0) {
+                        //Set D_i={l + 1, ..., l + expectedNumberOffspring}
+                        DsetCurrent = arma::linspace<arma::Col<unsigned int> >(l + 1, l + expectedNumberOffspring);
```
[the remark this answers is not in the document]
ilyaZar: Great catch!

inst/include/conditionalSampler.h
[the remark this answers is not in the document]
ilyaZar: Done!

ChangeLog
ilyaZar: Done!

adamjohansen:
I'm certainly not insisting on explicitly tidying up local variables, it just always seems to me that code is more readable if you get rid of things which you've explicitly created when you've finished with them and it keeps things tidy if the code is extended later (and I tend to distrust automatic garbage collection because I'm old...).

adamjohansen:
Time is finite and it may not be worth the energy, but the advantage of a linear Gaussian case is that you can compare Monte Carlo estimates with the truth which makes it easier to check that everything is working -- and particularly with the NC estimates, one can exploit unbiasedness to gain some confidence that everything is working.

adamjohansen:
I think we're probably ready to merge, but @LeahPrice raised the most recent queries so it probably makes sense for her to press the button if the current version look okay.

LeahPrice:
This all looks good to me now, thanks @ilyaZar for the hard work!

ilyaZar:
Thanks for merging @LeahPrice !

PR #67, opened 2021-09-11
Add testing and examples for conditional SMC.

ilyaZar:
So hopefully this goes into the right direction for testing and adding a simple example to consolidate some of the past work on conditional SMC. 

It's not yet ready for review but I decided to make it visible so you can have a look at it from time to time.


eddelbuettel:
Note that the CI failed here.  Maybe you forgot to commit an updated RcppExports.{R, cpp} or NAMESPACE file?

ilyaZar:
Yes I keep screwing things up quite strongly here for quite some time already ... 

I sorry I do not recall the exact workflow with `compileAttributes()` (used only the roxygen command from RStudio before).

If you have a minute, could you be so kind to  briefly sketch the workflow:

- add functions/files in `src/...` and `R/...`
- add documentation to `man/..`
- run `Rcpp::compileAttributes()` from within package folder to update `RcppExports.cpp` and `RcppExports.R`
- build the package

Now, `simGaussianSSM` is not documented within `./man` or `NAMESPACE` yet (which causes above CI error); but it doesn't need to be documented to be validly used within the package, or? (I only use it inside the function `compareNCestimates()` from the file ./R/cSMCexamples.R` and do not want to export it yet, still it seems to be alien to the package).

Just pushed again with updated `NAMESPACE` maybe that was the thing... I am probably missing something very obvious ?

eddelbuettel:
In the narrow sense, `compileAttributes()` deals with the Rcpp side.  It updates `RcppExports.{R,cpp}` as well as symbol registration for the dynamic library (otherwise in `init.c`). All is well there.  You can test that in baby packages created via `RcppArmadillo.package.skeleton()`.

Where users can tripped up is the R and `roxygen2` side. You either _do_ or _do not_ let `roxygen2` write `NAMESPACE`.  For many (smaller) packages I do not and then add the function to `NAMESPACE` myself.  Others, and I as well in larger packages, like `@export` tags in the roxygen header in the C++ source that `compileAttributes()` bring to R code (!!) from where `roxygen2` picks it up.  More brittle.

In any event, if it doesn't pass `R CMD build ...` followed by `R CMD check ...` locally then there is little point committing and just hoping it may sort itself out. It likely won't.   It can all be extra tedious but try to break it into small steps you can check one by one.

eddelbuettel:
BTW, given that we are all one happy SMC family now you can just branch off RcppSMC/rcppsmc and it becomes easier for us to follow.

ilyaZar:
Thanks for the information @eddelbuettel !

Yes keeping on and try to hammer through with my forced pushes, hoping for a miracle hole to appear in the CI wall, (which is rather well protected against brute force and should be passed via thinking) is unwise for sure...

I'll give it a try next week, we are not in a hurry (I hope).

eddelbuettel:
In this case you were missing the second closing `]` in `// [[Rcpp::export]]`.  The little things.  

(The fact that `compareNCestimates_imp` was not in either `RcppExports.*` was a tell.  Once I fix that I get ... a set of vanilla compiler error -- can I hand that back to you?)

ilyaZar:
Yes!! I'll check it locally, thanks for the hint!

ilyaZar:
__*Note for all:*__ though CI passed, this remains highly WIP and is far from ready for a merge since the other parts from #63 are meant to be implemented within this PR as well.

ilyaZar:
> BTW, given that we are all one happy SMC family now you can just branch off RcppSMC/rcppsmc and it becomes easier for us to follow.

Just tried that  @eddelbuettel but it seems I do not have write/read permissions (or I am doing sth. completely wrong here...)

```
git push -u upstream conditional-smc-testing 
ERROR: Permission to rcppsmc/rcppsmc.git denied to ilyaZar.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
```

I am absolutely fine with working with the fork and do PRs from there (as before), so no need to change this for me, but could probably make things easier for all to follow, as you said above.

eddelbuettel:
What does `git remote -v` say?  Do you by change still have the https:// entries?

```sh
edd@rob:~/git/rcppsmc(master)$ git remote -v
origin  git@github.com:eddelbuettel/rcppsmc.git (fetch)
origin  git@github.com:eddelbuettel/rcppsmc.git (push)
edd@rob:~/git/rcppsmc(master)$ 
```

ilyaZar:
Merged the changes from the other PR, so nothing to have a look at atm. Will try to fill in my additions over the weekend so we may have a more detailed look at it next week or so!

adamjohansen:
Thanks, @ilyaZar, indeed the probabilities need to be positive for it to make sense. I think the presentation in the technical report was a bit lazy (and should be fixed) and this expression applies to those _i_ for which the probability is non-zero and that the issue is resolved by replacing the probabilties specified here with the maximum of 0 and the expression given.
What you need to arrive at is a probability corresponding to the proportion of the region between j/n and (j-1)/n which the inverse of the empirical distribution function maps to index _i_.
Does that resolve the issue or have I missed something else? (Thanks for pointing this out, we'll tidy it up if that ancient document ever sees the light of day.)

ilyaZar:
I am still fixing the issue from back then regarding residual resampling, but some preliminary results are:

```
set.seed(123)
dat <- RcppSMC::simGaussianSSM(len = 20,
                               stateInit = 0)
out <- RcppSMC::compareNCestimates(dataY = dat$measurements,
                                   trueStates = dat$states,
                                   numParticles = 1e3,
                                   simNumber = 1e2,
                                   plot = TRUE)
```

produces
RplotNotWork

which shows that residual resampling is still incorrectly implemented.

However, we may have a look at the other resampling schemes (the plot of which I produced manually)
RplotWorks

Notably:

- the conditional and standard BPF likelihood estimates look similar if compared pairwise with the same resampling type, which is a good thing
- the kalman filter likelihood estimate is a bit off; it may be a bad data sample (seed is fixed or time series length is only 20); I am checking on it!

What is your feeling on this? Maybe some of my thoughts:

We may want to discuss if we would like to have all this plots jointly in a single big one or split them into smaller ones, or any other aesthic changes to the plots. 

Regarding the code I have the example implementation (`cSMCexamples.cpp`, `cSMCexamples.h` and `cSMCexamples.R`) coming along with some changes to the main library i.e. the conditional resampling facilities, and, in particular above fix (@adamjohansen thanks it worked! using max(previous_expression, 0) fixed the issue) in the file `conditionalSampler.h` where I also added some clearer documentation of the algorithms and refactored the code a bit for clarity.

We may want to split up these changes a bit, the PR is quite messy atm. because there is a lot of changes to different places. Also there will be another change to the conditional resampling facilities fixing the residual resampling bug, and another one to the `cSMCexamples.R` if there is something wrong with the Kalman likelihood estimate (I would tend to say yes here, since the red line in above plots seems to be too much off).

ilyaZar:
Updates the PR (plus I edited the above more detailed explanation).

in short:
there was an issue from conditional residual resampling which I (hopefully) fixed at the weekend (we excluded it from the main branch back then in GSOC); I think it looks far better now.

Otherwise the Kalman filter estimate (red line) is a bit off (wrote the code myself and its within the RcppSMC library, so there may actually be a mistake somewhere; I am checking on it atm!), but all the SMC log-likelihood estimates seem to have a similar range (which is probably a good thing).

LeahPrice:
I've taken a look at the output Ilya created here and in #63 and it all looks reasonable to me. I've taken a quick look at the code but haven't checked every line thoroughly. I suggest we merge it now as Ilya suggests so he can move on with other things. What do you think @adamjohansen and @eddelbuettel?

ilyaZar:
If all give green light (please feel free to add any suggestions before of course!!) I could go and merge this (would be my first merge, so I am curious to try).

Overusing the screenshot functions once more, write access is given here in this repo (otherwise I could not merge), but (and this is strange, or I am missing something obvious) the repo where this PR lives, https://github.com/rcppsmc/post-gsoc/pull/3, seems to not grant write access **_or_** I am not able to merge because of some Github quirk I am unaware of. 

See the two snippets

## From this repo:



## From https://github.com/rcppsmc/post-gsoc/

adamjohansen:
Sorry for being so very slow to get back to you with this, given the work involved I did want to take the time to have at least a bit of a look at it (and somehow although I kept expecting to have time to fit it in other things kept happening).

It looks great. I'd be happy to merge. I'll have a slightly more careful look at the conditional resampling code when things have settled down but it looked good on a quick reading and the testing results are very suggestive.

There's a trade-off between being self-contained and re-inventing the wheel and I'd have been inclined to use, say, `fkf::fks` to do smoothing for a linear Gaussian model rather than implementing something that's already been done in a fast and extensively tested way but it isn't a big deal with something this simple particularly given that we only need a univariate implementation (and I can see there are advantages to avoiding multiplying dependencies which are only used with what you might think of as rather ancillary code).

adamjohansen:
> Overusing the screenshot functions once more, write access is given here in this repo (otherwise I could not merge), but (and this is strange, or I am missing something obvious) the repo where this PR lives, https://github.com/rcppsmc/post-gsoc/pull/3, seems to not grant write access or I am not able to merge because of some Github quirk I am unaware of.

I've now added you as a contributor with write access to that repo. Hopefully that resolves the issue.

eddelbuettel:
Seconded.  It's 2022, we've been here for a while together now and training wheels should be off for everybody -- we may as well set everybody to owner.  Sadly, I think, changing it at the org level does not auto-propagate to repos within so if you "see something, say something" i.e. just ping @adamjohansen or myself and we adjust settings accordingly..

ilyaZar:
@adamjohansen added `FKF::fkf` and `FKF::fks` to the last commit  and removed the manual implementation.

1. I agree that there is a trade-off and though I checked my results against the `FKF` and both got the same likelihood values etc. it's probably for the better to rely on that package as also other users may recognize it.

2. Regarding code dependency/duplication: for a perfect world one might think about the "cleanest" way being with 2 "real" dependencies `Rcpp` and `RcppAramadillo` (`kitten` was not a strict dependency) and re-working all old examples and move them into some demo package (e.g. `RcppSMCexamples` or `RcppSMCclassics`) and polish their figures, documentation etc. 

That might have the appeal that it allows demo/example packages to be with loads of dependencies, nice graphics from `ggplot2`, the `FKF` or other implementation dependencies, or other data handling functions from the `tidyverse`, and god knows what, and if there is an issue with some dependency it is "only" the examples package that is affected whereas the main `RcppSMC` implementation still works perfectly due to its minimal dependencies. But that's just a thought and certainly something for the future, if it becomes important.

3. Also added the boxplot of the last commit to #63 for documentation purposes, but it all looks the same as expected.

@eddelbuettel @LeahPrice @adamjohansen I'll leave that PR for a couple of days if someone comes up with something, and if not, I'll be happy to merge it by the end of the week!

adamjohansen:
Looks good to me @ilyaZar and it doesn't hurt to have a few fewer lines of code to maintain.
On 2. I agree in principle separating essentials from examples which have their own dependencies does make sense in principle. (But time is finite and I don't think the current setup is causing significant difficulties at the moment.)

adamjohansen:
(So I think you can go ahead and merge, if that wasn't apparent in what I wrote...)

PR #72, opened 2022-04-19
Add RcppSMC package skeleton function

ilyaZar:
Sorry for this "out of the blue" pull request. No need for a quick answer/review. I just want to bring out this stuff so we can have a look at this from time to time and add this (or drop entirely) to the project and JSS paper in the future (this relates to the last chapter from the JSS paper-draft that is a bit unfinished and only has some drafts/ideas for a skeleton function).

The PR is a bit longer and certainly a side issue, but one that might be relevant in some future time to provide a skeleton function for convenient creation of packages that make use of `RcppSMC`.

Things done so far:
1. the main function is `RcppSMC::RcppSMC.package.skeleton()` which creates a new `R` package from scratch and automatically links to `RcppSMC`
2. it works very similar to `RcppArmadillo::RcppArmadillo.package.skeleton()` but adjusted to the `RcppSMC` use case
3. it exports as an example into the newly created package (via `.cpp` files and `.Rd` files, see content of `inst/skeleton`) so far just one `C++` function that makes use of `RcppSMC` for log sum weight computation: `NewPackageName::rrcppsmc_logNormWeightsCpp`
4. It exports as another basic example an R function `NewPackageName::hello()` similar to other packages that make use of skeleton function for pkg-creation (`RcppArmadillo` etc).

Things to consider:
- we might want to have some more examples, so far I used only one given in the JSS paper as a starting point


eddelbuettel:
In your sources. the directory `skeleton/` has to be `inst/skeleton/` so that it becomes installed as `skeleton/`.  A quick check against `Rcpp`, `RcppArmadillo`,  `RcppEigen`, `RcppGSL`, .... or any package using `inst/*` should convince. Very confusing the first time it bites but it is documented in WRE.

ilyaZar:
Thanks!
So technically it works meaning that I can create a package which makes use of `RcppSMC` via `RcppSMC:::RcppSMC.package.skeleton`.

It does not export the R-function directly into the package namespace (I will find that mistake eventually...) so `RcppSMC::RcppSMC.package.skeleton` is not working.

This new package builds/installs, passes `R CMD check` and the exported function `NewPackageName::rcppsmc_hello_world()` which makes use of the `RcppSMC` package computes log-weights and the log sum in line with what is in the JSS paper. The standard `hello_world`- R-function works too.

We may want to add better examples later etc. 

As written above, this is not urgent but looking forward for any improvements @eddelbuettel @adamjohansen @LeahPrice whenever you find the time to have a look!

eddelbuettel:
> so RcppSMC::RcppSMC.package.skeleton is not working.

I do not have the time to debug this for you but you _literally_ have four working implementations to study.  Please either make it work, or withdraw it.  If it doesn't work why do we have a PR?

ilyaZar:
Fixed the bug (forgot the `NAMESPACE` entry....) Sorry that was to fast forward; should have said that I am on it!

I'll stick to _draft PRs_ as we did before.

eddelbuettel:
Nice!  Also, and with apologies from here, these skeleton functions all have way too many arguments and options which nobody uses (_i.e._ `modules` for `Rcpp` and on).  So I'd be happy with a minimal simple one.  The point is to create a _simple_ working package. No more no less, and I find these help.  So you kicking this can down the road is great!

ilyaZar:
Thanks! 

I am not sure whether I should try to further simplify the skeleton-function or leave it as is?

If we are on the minimalistic road, then we could settle with just one example rpccsmc C++ function and one R-function (see my edited top level comment, last bullet point suggests to provide some more RcppSMC C++ functions but that might be not necessary anymore).

LeahPrice:
Thanks Ilya, it'll be nice to have this functionality. I like the idea of making it very simple and just having a single C++ function and a single R function in there. I don't think I'll have time to review the full code, at least not for another week or two. I can try to look at specific sections in a few days though. What would you like me to focus on?

ilyaZar:
Thanks @LeahPrice ! Atm I have no particular idea what could be reviewed, besides maybe the following part:

https://github.com/rcppsmc/rcppsmc/pull/72/commits/4315cdc780fdbd55105c568889e047aea9f9dfc2#diff-8d10b929ec72ca8253d3d9f8488f6126e80a31a036e957569a2138df96eeac56L81-R99

For the `Makevars` and `Makevars.win` files I copy paste from `RcppArmadillo` which I assumed to be installed since `RcppSMC` is likely to depend on Arma in the future. So changes to those files that occur in future upgrades to `RcppArma` automatically carry over to `RcppSMC`; which one may or may not like.

The other choice would be to have individual `Makevars` and `Makevars.win` inside `rcppsmc/inst/skeleton/...` and manually update the content when necessary.

I am not sure which way is smarter in the longer term. Maybe it's really not that crucial, too.

eddelbuettel:
I had never thought of the skeleton generator copying a dependencies from another package -- but you have a point here.  In general we kept local reference copies but give that RcppSMC cannot live without RcppArmadillo we may as well take that. 

Either was not a biggie as this Makefile stanza changes little.    And yes, simpler is better. :)

ilyaZar:
Another thing @eddelbuettel . I just checked f `RcppSMC::RcppSMC.package.skeleton()` and figured out that it fails __*whenever*__ `pkgKitten` is installed (because it calls `kitten()` instead of `utils::package.skeleton()`) in two ways:

1. if `force = TRUE`, it does not overwrite the package directory but produces an error
2. the other argument `code_files` produces an error

`RcppSMC::RcppSMC.package.skeleton()` is pretty much the same as `RcppArmadillo::RcppArmadillo.package.skeleton`, `RcppEigen::RcppEigen.package.skeleton` etc. so these functions are "affected" (maybe this is not relevant to care about at all) as well.

In both cases "Calling kitten to create basic package" is part of the error message for 1. and 2. suggesting that `kitten()` is called under the hood, but the help files in all skeletons (SMC, Arma, Eigen, and probably the other GSL etc.) do not mention this so in all cases one has to read the function body (maybe that's how life is sometimes).

Is there a particular reason why `force` and `code_files` are not part of `pkgKitten::kitten()`?

I am asking this because either I could try to fix/improve 1. and 2. for `RcppSMC`-package skeleton and one could carry this over to Arma and Eigen versions.

But a nicer way (solving all at once) could be to just update `pkgKitten::kitten()` with support for `force` and `code_files`, if you seem this relevant for the kitten package? (could be too much/unnecessary though).

eddelbuettel:
I don't know off hand and don't have time to chase it.  I did make some changes to `pkgKitten` and how it can be used with the other packages, so on the margin we want what it does for `RcppArmadillo`.   And I strongly recommend to not escalate.  Please do not go off modifying / PRing for other packages.

Quick suggestion: can you set up `RcppSMC.package.skeleton` to simply not have arguments `force`, `code_files`, `example_code` ?

adamjohansen:
Thanks for this, @ilyaZar and sorry for being slow to respond.

Superficially this looks like a nice addition, modulo complications with getting it to play nicely with other packages. I won't have a chance to look at this in any detail before the weekend; if there is anything specific that I might be able to contribute then do let me know -- I can't claim any particular expertise in this area.

I would say that if we are still planning on writing the JSS thing then we probably ought to focus on that, and any code that's essential to getting it finished rather than worrying too much about things that are "nice to have" because time is limited and the main barrier to entry with this library at present is, at least IMHO, the lack of proper documentation.

ilyaZar:
@adamjohansen totally agree, I'll finish #67 and then we can turn to the paper.

I do not think that there is something important left to do for you or @LeahPrice except maybe to go through few lines of the `ChangeLog` file from the last commit in case of typos etc.

@eddelbuettel as you requested, I set up `RcppSMC.package.skeleton` without 
`force`, `code_files`, but are you sure about `example_code` (from the help file):

```
If TRUE, example c++ code using RcppArmadillo is added to the package
```

I'd be inclined to keep it because we wanted to have the example RcppSMC `C++` code or that not being generated for more advanced users.

eddelbuettel:
Maybe remove the option but always copy one example in?  That is what RcppArmadillo for probably a decade (with the outer-and-inner product of the column vector given).

ilyaZar:
Why not simply keep it, it does not cause an error (only `force` and `code_files` did), sorry I seem to be confused here...

eddelbuettel:
With default value `TRUE` --> that would amount to the same.  Fine by me.   In a better world we'd also check what happens with `example=FALSE` runs.

Please also make sure you update the help text to `RcppSMC` as what you quoted a few message above still references the other package.

ilyaZar:
Hi Dirk, yes you are absolutely right but I think it's "O.K" because for `example_code = FALSE` _the exact same behavior_ for `RcppSMC`, `Arma`, `Eigen` and all of these is triggered.

You asked me not to escalate and to PR anything so I pretty much stopped my investigations here; I am fully aware that `R CMD check` will not pass since no `.cpp` or `.Rd` files are added,  `Rcpp::compileAttributes()` is never called and `R CMD check` will complain about a missing `exports.cpp` or `.so` files or sth. like that.

Maybe it's fair to say that a reasonable user setting `example_code=FALSE` should be able to fix `R CMD check` after adding code/`.cpp`-files and call `Rcpp::compileAttributes` manually.

I added a note to the help text of that argument in case of `example_code = FALSE` and checked the wrong reference.

eddelbuettel:
Please see https://github.com/RcppCore/RcppArmadillo/blob/master/R/RcppArmadillo.package.skeleton.R#L24 and realize that adding example code is the default, leading to a working -- and _or course_ zero WARNING or NOTE -- package. Why else would we do this?

eddelbuettel:
Yes for _existing_ packages it is painful to remove _existing_ argument because changing the interface is bad.

You are proposing a new one.  It is has a non-optional argument (that really only works set as TRUE).  So I suggested to remove it (and of course to keep the bahviour).  I probably didn't explain that clearly enough over the previous dozen messages.

ChangeLog
```diff
@@ -1,3 +1,16 @@
+2022-04-22  Ilya Zarubin  <ilya.a.zarub@gmail>
+
+	* R/RcppSMCpackageSkeleton.R: add skeleton function for straightforward
+	package creation of external packages that make use of RcppSMC; generates
+	minimal boiler-plate package to enable the user an easy start and interface
+	with RcppSMC
```
eddelbuettel, line 6: Please reduce to a line, or two at most.

ChangeLog
```diff
@@ -1,3 +1,16 @@
...
+	* man/RcppSMCpackageSkeleton.Rd: add skeleton function documentation
+
+	* inst/skeleton/rcppsmc_hello_world.cpp: C++ example function to compute
+	sum of logarithmic particle weights; illustrates how to write user-defined
+	C++ functions that make use of RcppSMC functions in a separate package
```
eddelbuettel, line 11: Recommend to remove everything past the semicolon in line 10

R/RcppSMCpackageSkeleton.R
```diff
@@ -0,0 +1,92 @@
+RcppSMC.package.skeleton <- function (name = "anRpackage", list = character(),
+                                      environment = .GlobalEnv,path = ".",
+                                      example_code = TRUE) {
```
eddelbuettel, line 3: Can we remove `example_code` as discussed?

R/RcppSMCpackageSkeleton.R
```diff
@@ -0,0 +1,92 @@
...
+    }
+    src <- file.path(root, "src")
+    if (!file.exists(src)) {
+        dir.create(src)
+    }
+    man <- file.path(root, "man")
+    if (!file.exists(man)) {
+        dir.create(man)
+    }
+    skeletonArma <- system.file("skeleton", package = "RcppArmadillo")
```
eddelbuettel, line 64: Do we need Armadillo here?

R/RcppSMCpackageSkeleton.R
```diff
@@ -0,0 +1,92 @@
...
+    skeletonSMC  <- system.file("skeleton", package = "RcppSMC")
+    Makevars <- file.path(src, "Makevars")
+    if (!file.exists(Makevars)) {
+        file.copy(file.path(skeletonArma, "Makevars"), Makevars)
+        message(" >> added Makevars file with Rcpp settings")
+    }
+    Makevars.win <- file.path(src, "Makevars.win")
+    if (!file.exists(Makevars.win)) {
+        file.copy(file.path(skeletonArma, "Makevars.win"), Makevars.win)
+        message(" >> added Makevars.win file with RcppArmadillo settings")
```
eddelbuettel, line 74: Why are we referencing settings from another package?

R/RcppSMCpackageSkeleton.R
```diff
@@ -0,0 +1,92 @@
...
+    }
+    if (example_code) {
+        file.copy(file.path(skeletonSMC, "rcppsmc_hello_world.cpp"),
+            src)
```
eddelbuettel, line 78: We have no coding conventions that insists on 80 columns.  I'd wrap src to the previous line.

R/RcppSMCpackageSkeleton.R
```diff
@@ -0,0 +1,92 @@
...
+    root <- file.path(path, name)
+    DESCRIPTION <- file.path(root, "DESCRIPTION")
+    if (file.exists(DESCRIPTION)) {
+        x <- cbind(read.dcf(DESCRIPTION),
+                   Imports = sprintf("Rcpp (>= %s)",
+                   packageDescription("Rcpp")[["Version"]]),
+                   LinkingTo = "Rcpp, RcppArmadillo, RcppSMC")
+        write.dcf(x, file = DESCRIPTION)
+        message(" >> added Imports: Rcpp")
+        message(" >> added LinkingTo: Rcpp, RcppArmadillo")
```
eddelbuettel, line 42: We also add RcppSMC

R/RcppSMCpackageSkeleton.R
```diff
@@ -0,0 +1,92 @@
...
+    man <- file.path(root, "man")
+    if (!file.exists(man)) {
+        dir.create(man)
+    }
+    skeletonArma <- system.file("skeleton", package = "RcppArmadillo")
+    skeletonSMC  <- system.file("skeleton", package = "RcppSMC")
+    Makevars <- file.path(src, "Makevars")
+    if (!file.exists(Makevars)) {
+        file.copy(file.path(skeletonArma, "Makevars"), Makevars)
+        message(" >> added Makevars file with Rcpp settings")
```
eddelbuettel, line 69: Why are we referencing Rcpp?  (Even if the other package does, it is still silly as we are about RcppSMC here)

inst/skeleton/rcppsmc_hello_world.cpp
```diff
@@ -0,0 +1,23 @@
...
+//
+// via the exports attribute we tell Rcpp to make this function
+// available from R
+//
+// [[Rcpp::export]]
+Rcpp::List rcppsmc_logNormWeightsCpp(arma::vec logWeights){
+  double logSum = smc::stableLogSumWeights(logWeights);
+  logWeights -= logSum;
+  return Rcpp::List::create(Rcpp::Named("logSum") = logSum,
+                            Rcpp::Named("logNormWeights") = logWeights);
```
eddelbuettel, line 22: If the results are both scalars it it easier to return a numeric vector of length 2?

eddelbuettel (commented):
There are a few things I'd still change.

R/RcppSMCpackageSkeleton.R
ilyaZar: I changed to "added Makevars file for baseline compiler flags" and "added Makevars file for baseline compiler flags(windows)" so the user of the skeleton gets informed why.

R/RcppSMCpackageSkeleton.R
ilyaZar: "added Makevars file for baseline compiler flags(windows)" , see above

R/RcppSMCpackageSkeleton.R
ilyaZar: Yes, for copying the `Makevars` and `Makevars.win`.

inst/skeleton/rcppsmc_hello_world.cpp
ilyaZar: @eddelbuettel that's tricky. Essentially this is an example from the JSS paper, which is likely to change anyway. I wanted to have consistency between what the skeleton does ("exact" line by line) and what the JSS will provide as an example, mostly for clarity purposes.

SInce this is likely to change anyway, as we might update the example from the paper, I can of course change it __*now*__  but it it might change in the future anyway. What do you think?

ilyaZar:
Thanks for the comments @eddelbuettel , all except for the last comment, ready I hope!

R/RcppSMCpackageSkeleton.R
eddelbuettel: Yes, as discussed.  It's doable, but still a bit weird.  Either way is fine though.

eddelbuettel:
If you run your branch with `R CMD check --as-cran` as I would assume you did already you see (here I used the `rcmdcheck::rcmdcheck` wrapper hence the little icons)

```r
✔  checking loading without being on the library search path ...
W  checking dependencies in R code ...
   '::' or ':::' import not declared from: ‘pkgKitten’
   'loadNamespace' or 'requireNamespace' call not declared from: ‘pkgKitten’
✔  checking S3 generic/method consistency (537ms)
```

and

```r
✔  checking foreign function calls ...
N  checking R code for possible problems (2s)
   RcppSMC.package.skeleton: no visible binding for global variable
     ‘package.skeleton’
   RcppSMC.package.skeleton: no visible global function definition for
     ‘packageDescription’
   Undefined global functions or variables:
     package.skeleton packageDescription
   Consider adding
     importFrom("utils", "package.skeleton", "packageDescription")
   to your NAMESPACE file.
✔  checking Rd files ...
```

Can you please fix this?

ilyaZar:
Added `Suggests: pkgKitten` and imported the two missing funcs via `NAMESPACE`.

Two small follow up questions:

1. In the `DESCRIPTION` file I saw:
`Imports: Rcpp (>= 0.11.0), methods`
is that strictly necessary to add `methods` here since it is part of `base`, or?

2. What version number would you suggest for `pkgKitten`; I've somewhere read that one should (or it is at least a good habbit/ curtosy to the user) to always add a version number to something mentioned in `Imports:` or `Suggests:`. And if one is not sure about which version number, then being conservative and adding my platform version number is the way to go?

eddelbuettel:
Re 1:   I don't remember the details anymore but as some point we _had to_ load `methods` for S4 methods because even though R always _installs it_ (as it comes with R) it didn't always load it.   That may have changed.   In any event, this a) does no  harm and b) `R CMD check --as-cran` does not complain both of which are good enough for me, and I hope you can live with them too.

Re 2:  I do not agree that every suggestion or dependency should versioned.  If can't actually think of a minimally required version, why would you impose one?

eddelbuettel:
Thanks for making the changes!  I think I will merge now and as the commit history had got a little disjointed I suggest to merge as a 'squash merge' which makes it one big commit -- is that ok with you?

Good to have this now.

eddelbuettel:
Done -- and rolled to 0.2.6.1 to differentiate from the last CRAN version.

